PR #858, opened 2015-01-28
Remove the netlib python module from the pathod package

afics:
The pathod release tarballs include the python2-netlib module which we
have packaged separately.


afics:
Resolved upstream.

PR #888, opened 2015-04-10
[pcredz] Fix dependency error

man0lis:
pcredz actually uses the pylibpcap package instead of pypcap. Without this change pcredz will refuse to parse any files


noptrix:
thx

PR #895, opened 2015-04-26
Fix dependencies on cangibrina

fnk0c:
Added 'python2-beautifulsoup4' 'python2-socks' 'python2-mechanize' as dependencies


noptrix:
thx

PR #907, opened 2015-05-19
Fix extremely dangerous failing post-install script.

ahills:
The strangely named /var/lib/nsia/var directory was not created by the PKGBUILD's package() function, and the post-install script runs recursive group and mode changes on the root directory, destroying existing installations. These changes correct this oversight.


noptrix:
thx

PR #915, opened 2015-07-04
Updated metasploit dependencies

jfpdx:
Metasploit binary package fails to install correctly unless libxslt is present, added the dependency to PKGBUILD.


noptrix:
thx a lot

PR #922, opened 2015-07-17
winexe: fix for gnutls > 3.0.0

Rogdham:
Winexe did not work since gnutls versions > 3.0.0.
Also some perl problems with `defined`.

Please note that Sourceforge is currently having some issues, so you may need to download the source by hand.


noptrix:
thanks a lot

PR #932, opened 2015-08-12
recon-ng upgraded to version 4.6.3

ChrisFernandez:
I was trying to run today recon-ng on my own blackarch system for a pentest and notice it was outdated and was not able to run. so forked the project and did the changes for me locally, then i decided to push the change upstream so other people can also use it.


noptrix:
thx a lot

PR #939, opened 2015-08-21
airflood 

dwaddle:
Changed download location to working one. Replaced sha1sum verification by md5sum the md5sum is provided by the download site


noptrix:
thx. though, we use sha1sums not md5. i will change it

PR #965, opened 2015-10-28
Fix `elite-proxy-finder` missing dependency

nil0x42:
The `python2-beautifulsoup3` dependency was missing in this package


nil0x42:
edit: forgot to change pkgrel, will do another pull request

PR #989, opened 2015-12-22
Add missing dependency on 'sipsak' package

nil0x42:
## Fix missing 'c-ares' dependency on 'sipsak' package

**Without c-ares package installed, running sipsak produced the following error:**
```
$ sipsak
sipsak: error while loading shared libraries: libcares.so.2: cannot open shared object file: No such file or directory
```


noptrix:
thank you.

PR #997, opened 2016-01-07
Update PKGBUILD

NovaCygni:
Updated V3n0M Scanner

### DIFF
--- a/packages/v3n0m/PKGBUILD
+++ b/packages/v3n0m/PKGBUILD
@@ -1,26 +1,23 @@
 # This file is part of BlackArch Linux ( http://blackarch.org ).
 # See COPYING for license details.
 
+# This file is part of BlackArch Linux ( http://blackarch.org ).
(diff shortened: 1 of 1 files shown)


noptrix:
thanks. i have updated and released the scanner. but i didn't use your pkgbuild file, because it's wrong.  your code has inconsistency when it comes to python 2 and 3. also, it seems your setup.py is not really finished so i had to install manually. have a look in our PKGBUILD file for the scanner. cheers

noptrix:
https://github.com/BlackArch/blackarch/blob/a68c579051662c539219a0e821e425815ee2263a/packages/python-asyncio/PKGBUILD

noptrix:
ignore the previous link. here is the link to v3n0m: https://github.com/BlackArch/blackarch/blob/a68c579051662c539219a0e821e425815ee2263a/packages/v3n0m/PKGBUILD

PR #1001, opened 2016-01-12
All code upgraded to work properly with Python3.

NovaCygni:
All features now work as intended plus additions.

In this request Ive only changed the PKGBUILD To reflect the version number of the release.


noptrix:
already released ;)

PR #1070, opened 2016-06-08
ivre: version bump, use version number from setup.py, add opt dep

p-l-:
IVRE includes a [new feature for flow analysis](cea-sec/ivre#227) which comes with an optional dependency.

Also, the package version now includes IVRE's base version (0.9.3 for now). I'm not sure if it matches BlackArch versions policy, so please let me know if you prefer something else.

### DIFF
--- a/packages/ivre/PKGBUILD
+++ b/packages/ivre/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname='ivre'
-pkgver=898.078fc2d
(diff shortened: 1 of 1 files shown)


p-l-:
Conflict fixed after c8c6e2b0339ee33b321289e10f9a114d15f7e65f.

p-l-:
Maybe you could elaborate on that @noptrix?
- What would be the best way to include IVRE's version number?
- Why don't you want to add the optional dependency?

PR #1074, opened 2016-06-14
version bump of perl-io-socket to 1.36_01

gratuxri:


noptrix:
thx. will be released now

PR #1083, opened 2016-06-19
recon-ng version bump to 4.8.0

ChrisFernandez:
Hello again :) sorry to keep sending pull request for different packages, I work as a pentester and when some packages get behind I have to modify the package to install it.. in some cases like recon-ng I can't even use it unless I update, so figure to sent upstream my changes, i'm sure if there is other professional pentesters using blackarch they will be happy to see the version bump. Cheers.


noptrix:
thanks again

PR #1093, opened 2016-07-08
bump faraday to 2546.508a848

ChrisFernandez:
This is important since the new version defaults to GTK, the old one that is installed by blackarch defaults to QT and the UI comes out with international chars not western so is broken.. I just tested this build on my system and defaults to GTK and there is no issues so far.. let me know if you need anything else. cheers.


noptrix:
already done. thx for reporting

PR #1107, opened 2016-08-07
Added wxhexeditor-git

vdun:


noptrix:
thanks a lot. will be renamed to 'wxhexeditor' though as we don't use -git names. cheers for the pkgbuild

PR #1159, opened 2016-09-02
dump1090: Fixed public_html location

J08nY:
Fixes the location of public_html directory in `dump1090` so that it can show the map correctly when run with `--net`.


noptrix:
thank you

PR #1167, opened 2016-09-11
Vcsmap version bump

melvinsh:
Thanks for including vcsmap in BlackArch!

Today I released a new version that fixes a change in GitHub. This is an important update. Because I noticed that vcsmap is included in BlackArch I also made it easier to find the right directory (see: https://github.com/melvinsh/vcsmap/commit/6fba205d8d8fe32365775c32373d135cc6c4cf82). 

This is my first edit to a `PKGBUILD`, so I hope I got it right.


noptrix:
thank you.

PR #1177, opened 2016-09-21
update url and pkgdesc

amuntner:
Updating url and package description


heichblatt:
Builds just fine. Thanks @amuntner !

PR #1267, opened 2017-01-02
File conflicts

lsfxz:
While mucking about with my ruby-PR I've stumbled upon a few other packages which caused file conflicts and fixed those.


noptrix:
thanks a lot. we would like to merge in your PR. can you just synchronize before we do the merge?

noptrix:
thanks a lot

PR #1271, opened 2017-01-06
Initial attempt to fix veil

### LINKED ISSUE #1268: veil ??
whats the poing of having veil evasion when it doesnt have support && doesnt work ?

lsfxz:
I've started going through the open issues. This should (hopefully) fix #1268 .

There'll probably still be issues with wherever veil uses go. I'll look into this soon – I've (yet) only tested if it actually properly starts (and the wine-environment gets setup more or less as it should).


noptrix:
thx a lot

PR #1304, opened 2017-01-28
IVRE: use correct version, version bump, add optional dep

p-l-:
This adds an optional dependency and uses the correct version, based on IVRE's `setup.py --version` output.

### DIFF
--- a/packages/ivre/PKGBUILD
+++ b/packages/ivre/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname='ivre'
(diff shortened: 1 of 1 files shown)


noptrix:
thanks, this has been fixed manually without this PR.

p-l-:
@nopper this has not been fixed.

noptrix:
@p-l- ok, will be fixed

noptrix:
everything fixed + missing deps. the python --version makes an invalid version number format for PKGBUILD...so we keepthe current format git head etc.

PR #1322, opened 2017-02-08
new tool: dradis-ce

lsfxz:
I've finally found some time to look at the non-legacy dradis, dradis-ce.

To make it usable without running rails as root a few files and folders in `/usr/share/dradis-ce`  are made world writable, which is probably not the best way to approach this – I'm open to suggestions for better ways to handle this situation: maybe by creating a dradis-user who can write in those folders and launching dradis/rails with a service?


noptrix:
thanks a lot :) about the dir permissions, we can find a solution later by simply adding a user

noptrix:
added and released.

PR #1524, opened 2017-02-22
small dependency fixes

karlguy:
Wifiphisher now has all dependencies required. Also, I couldn't help but notice that python2-jinja2 was just a duplicate of python2-jinja, and this would lead to installation failures when trying to install both. So I deleted python2-jinja2, and changed all the necessary dependenceis


noptrix:
thanks a lot.

Issue #1551, opened 2017-03-06
[Error] Package analyzepesig


philosophos:
`pacman -Sg 'blackarch-windows blackarch-binary'`will show `analyzepesig`,
The package should be in group `blackarch-windows` and `blackarch-binary`,
not `'blackarch-windows blackarch-binary'`

noptrix:
fixed

PR #1552, opened 2017-03-06
Update analyzepesigs group information

ikstream:
This will handle BlackArch/blackarch#1551


noptrix:
thx!

Issue #1556, opened 2017-03-08
[Error] Install on top of Arch doesn't work


aeddi:
Hello,
I'm trying to install BlackArch on top of Arch but pacman abort installation due to conflicting files.
I tried to delete the problematic packages and then to change mirrors (4 times) before restarting the installation but it doesn't work either.

```
(2943/2943) checking for file conflicts            [####################] 100%
error: failed to commit transaction (conflicting files)
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/CHANGELOG exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/LICENSE exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/README.md exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/Rakefile exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/colorize.gemspec exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/lib/colorize.rb exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/lib/colorize/class_methods.rb exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/lib/colorize/instance_methods.rb exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/lib/colorized_string.rb exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/gems/colorize-0.8.1/test/test_colorize.rb exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.4.0/specifications/colorize-0.8.1.gemspec exists in both 'bettercap' and 'ruby-colorize'
/usr/lib/ruby/gems/2.3.0/gems/base32-0.3.2/.gitignore exists in both 'ruby-base32' and 'ssrf-proxy'
[...]

PR #1557, opened 2017-03-09
mutagen dependency of mat

lsfxz:
It seems like upstream extra/mutagen actually is python2-mutagen. Bad naming, I'd say.

Anyway, changing the dependencies of mat makes python2-mutagen obsolete and should fix some of the file conflicts of #1556.

The leftover conflicts.. well, you know where my PR is ;)


noptrix:
thanks a lot

Issue #1533, opened 2017-03-01
[Error] Package Wifi-Pumpkin


MatheusViGo:
When executed show this error message:
"""Traceback (most recent call last):
  File "wifi-pumpkin.py", line 45, in <module>
    from core.main import Initialize
  File "/usr/share/wifi-pumpkin/core/main.py", line 31, in <module>
    from core.widgets.tabmodels import (
  File "/usr/share/wifi-pumpkin/core/widgets/tabmodels.py", line 7, in <module>
    from core.utility.threads import ThreadPopen
  File "/usr/share/wifi-pumpkin/core/utility/threads.py", line 19, in <module>
    from core.servers.proxy.controller.handler import MasterHandler
  File "/usr/share/wifi-pumpkin/core/servers/proxy/controller/handler.py", line 1, in <module>
    from plugins.extension import *
  File "/usr/share/wifi-pumpkin/plugins/extension/js_inject.py", line 2, in <module>
    from mitmproxy.models import decoded
ImportError: No module named mitmproxy.models"""

#Repository updated and all dependencies satisfated

noptrix:
this is weird. i don't get this error on 2 different systems here. do you have more info on this?

MatheusViGo:
I installed on top of arch linux.

ikstream:
I installed it on a clean vagrant box and the error message came up as well.
I had to install mitmproxy with `pip2 install mitmproxy`
in addition i had to install 'python2-configobj' and  'python2-netfilterqueue'.
On the vagrant box it worked, but on my machine i ran into another error.

MatheusViGo:
I installed 'python2-configobj', 'python2-netfilterqueue' and mitmproxy with all dependecies satisfated and the error still persists.

ikstream:
Did you install mitmproxy with pip?
The regular install with pacman did not work for me either.

MatheusViGo:
Yes, I've successfully installed mitmproxy with pip, but now when I run wifi-pumpkin this error appears.
![captura de tela_2017-03-09_13-08-32](https://cloud.githubusercontent.com/assets/17281717/23758870/d70e1a80-04c9-11e7-8d6b-a5753b5aafa9.png)

ikstream:
That is the same problem i encountered on the VM, but i had no time to look into it by now.
Could you try this solution?
https://github.com/P0cL4bs/WiFi-Pumpkin/issues/53

MatheusViGo:
[...]

PR #1559, opened 2017-03-09
Add dependencies for wifi-pumpkin

ikstream:
The Problem i see here is, that if pip is used to install dependencies, they wont be removed on uninstall.
This adresses BlackArch/blackarch#1533


noptrix:
thx for this!

Issue #1549, opened 2017-03-04
[Error] Insanity doesn't work.


Y0lan:
The insanity bin was made to work for deb, because it contain deb command like apt-get etc... 
So if i try to launch it, it works but none of the option works.

PR #1586, opened 2017-04-03
Insanity - Set python2.7 as fix version in settings.py

ikstream:
This fixes BlackArch/blackarch#1549


noptrix:
thanks a lot

Issue #1566, opened 2017-03-14
Sparta not working in archlinux


g0rwx:
$ sparta
[-] Import failed. PyQt4 library not found. 
Try installing it with: apt-get install python-qt4

aancw:
Hi, try to install this dependency:

python2-elixir
python2-lxml
python2-pyqt4

Reference: https://aur.archlinux.org/packages/sparta/

Need to update PKGBUILD.

aancw:
Hm... Nope. From official repo they say that you need to downgrade pyqt4 version:

https://github.com/SECFORCE/sparta/issues/59

or using this: 

https://stackoverflow.com/questions/37604532/python-2-7-11-importerror-cannot-import-name-qtwebkit-kali-linux-debian-8

g0rwx:
I tried everything.

But it failed.

saltasatelites:
Dirty solution in Archlinux.
Works for me.
Go to the sparta directory
Open sparta.py.  In line 22 delete QtwebKit at the end of the line. Only QtwebKit (don 't forget to delete the last ","
Save changes
Open /sparta/ui/view.py
Delete completly lines 16, 50, 51, 52, 116, 423, 424
Save changes

Thts all. Hope it helps you!!!!!

NovaCygni:
saltasatelites Solution confirmed working, if not a litte dirty as described :/ but, yeh, it works.

g0rwx:
my log..

Traceback (most recent call last):
  File "sparta.py", line 88, in <module>
    view = View(ui, MainWindow)						# View prep (gui)
  File "/usr/share/sparta/ui/view.py", line 39, in __init__
    self.startConnections()								# signal initialisations (signals/slots, actions, etc)
  File "/usr/share/sparta/ui/view.py", line 120, in startConnections
    self.connectHelp()		
  File "/usr/share/sparta/ui/view.py", line 429, in connectHelp
    self.ui.menuHelp.triggered.connect(self.helpWidget.show)
AttributeError: 'View' object has no attribute 'helpWidget'

aancw:
@saltasatelites This will be good if you provide diff file for comparing :+1:

noptrix:
hey guys. if the solution @saltasatelites mentioned works, i will provide a patched version and release it.

NovaCygni:
Working here with his solution.

ChrisFernandez:
I have a diff issue.. says cant find  
```
[...]

PR #1587, opened 2017-04-04
Sparta dirty Fix

ikstream:
Add elixir as dependency and apply the fix @saltasatelites uses in BlackArch/blackarch#1566


noptrix:
thanks a lot

Issue #1547, opened 2017-03-04
[Error] Package Airgraph-ng 


MatheusViGo:
When executed show this error message:
"""Traceback (most recent call last):
  File "/usr/bin/airgraph-ng", line 17, in <module>
    from graphviz import *
  File "/usr/lib/python2.7/site-packages/graphviz/libOuiParse.py", line 36, in <module>
    raise Exception("Could not determine path, please, check your installation")
Exception: Could not determine path, please, check your installation"""

#Repository updated and all dependencies satisfated

aancw:
Looking at airgraph-ng source will be like this:

```
if os.path.isdir('./support/'):
    path='./support/'
elif os.path.isdir('/usr/local/share/airgraph-ng/'):
    path='/usr/local/share/airgraph-ng/'
elif os.path.isdir('/usr/share/airgraph-ng/'):
    path='/usr/share/airgraph-ng/'
else:
    raise Exception("Could not determine path, please, check your installation")
```

So, if you create new directory location from above, it will be working well. 

cc @noptrix

PR #1588, opened 2017-04-04
Airgraph-ng create directory

ikstream:
Create a directory as @aancw recommends in BlackArch/blackarch#1547


noptrix:
thanks a lot

PR #1599, opened 2017-04-11
update to veil 3

### LINKED ISSUE #1598: [OUTDATED PACKAGE] Veil Framework
The veil framework has completely changed now.

The original github page (https://github.com/Veil-Framework/Veil) says:
    Veil Evasion is no longer supported, use Veil 3.0!

New repository is located here:
    https://github.com/Veil-Framework/Veil

lsfxz:
So I tried to get this working again. Man, I don't like its install process..

Anyway: As before Go payloads currently fail. I'll see that I'll actually get this done somehow this time, but not today, I guess. It would be nice to get this working without using gcc-go, which conflicts with gcc..
Also during Ruby payload generation ocra semi-silently fails:
```
C:/Ruby187/lib/ruby/gems/1.8/gems/ocra-1.3.6/bin/ocra:1040:in `initialize': Permission denied - tmpin (Errno::EACCES)
```
..even though directly executing _exactly_ the os.system call veil uses doesn't fail. I'm not sure if that's an issue with my kernel (grsec) and wine or something else – if anyone feels like testing or has an idea I'm all ears.

Should (somewhat) fix #1598

### DIFF
--- a/packages/veil/PKGBUILD
+++ b/packages/veil/PKGBUILD
@@ -2,26 +2,28 @@
 # See COPYING for license details.
 
 pkgname='veil'
-pkgver=602.3c0dec3
-pkgrel=1
-epoch=2
+pkgver=67.7b22db7
+pkgrel=3
+epoch=3
 groups=('blackarch' 'blackarch-automation' 'blackarch-exploitation')
 pkgdesc='A tool designed to generate metasploit payloads that bypass common anti-virus solutions.'
-url='https://github.com/veil-evasion/Veil'
+url='https://github.com/Veil-Framework/Veil'
 arch=('x86_64' 'i686')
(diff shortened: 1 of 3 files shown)


lsfxz:
Ok, actually func_go_deps doesn't even do what it says it does (build go). Neither does it require gcc-go.  But it wants to delete `/usr/bin/go`. Yeah, nope.
If you instead limit it to just extracting go to `/usr/local/go` everything seems to work, payloads get built as the should.
As good a tool it might be, but that setup.sh needs to be burnt and redone from scratch, it seems.

noptrix:
@lsfxz mate, really thanks a lot. appreciated for this!

noptrix:
yeah, the install proc. sucks.

PR #1604, opened 2017-04-14
fix dradis-ce permission issues

evs-ch:
I finally got around to fixing the permission issues dradis-ce would have, by adding a dradis-ce group and making /usr/share/dradis-ce group-writable to anyone in that group.


noptrix:
thanks a lot

noptrix:
@evs-ch i just changed the style in PKGBUILD a bit and also set pkgrel to '2' as we were on '1' before.

PR #1730, opened 2017-07-16
Add New Tool : CyberScan

medbenali:
CyberScan is able to send and capture packets of several protocols, forging and decoding them to be used to most network tasks such as scanning, pinging, probing, and attacks.

https://github.com/medbenali/CyberScan

I have the honor to add it in Black Arch


noptrix:
thanks

PR #1734, opened 2017-07-31
Update sylkie PKGBUILD

dlrobertson:
- Add the [ENABLE_SETUID](https://github.com/dlrobertson/sylkie/wiki#build-options) build option
 - Move libseccomp from makedepends to depends. We don't statically link
   against it.


noptrix:
thanks for this!

PR #1742, opened 2017-08-12
[update] sploitctl has now a bash-completions.

PedroSFreitas:
This pull is related to https://github.com/BlackArch/sploitctl/pull/8 in sploitctl.

I'm not sure about `pkgver`, so a review would be interesting.


noptrix:
thanks a lot

PR #1743, opened 2017-08-12
Added libxslt as a dependency for beef

MD5HashBrowns:
Beef wouldn't build without `libxslt` because `nokogiri` is being built with system libraries.


PedroSFreitas:
I'm getting a weird problem running this `PKGBUILD` file.

On my BlackArch Live ISO, hacked to run `makepkg` with `root` I get:
BlackArch

On my Archlinux normal installation (workstation), I get:
ArchLinux

Can you build this package without problem?
Maybe I'm missing something. :cry:

noptrix:
thank you

PR #1747, opened 2017-08-21
Added new tool massbleed

ChrisFernandez:
Hey guys I notice the tool sn1per in the blackarch repos needs this tool as well. since is not in the repo nor in arch repos here I created a simple install.
keep it up.


noptrix:
@ChrisFernandez thank you!

PR #1753, opened 2017-08-24
Another s1per dependency called goohak

ChrisFernandez:
Found missing another dependency on sn1per:
Sn1per/sniper: line 1021: goohak: command not found
so created the package, cheers.


noptrix:
thanks a lot

Issue #1697, opened 2017-06-20
[Tool request] rtfm


vdun:
https://github.com/leostat/rtfm

PR #1758, opened 2017-08-31
[tool request] adding rtfm #1697

PedroSFreitas:
https://github.com/BlackArch/blackarch/issues/1697

image

image

### DIFF
--- a/packages/rtfm/PKGBUILD
+++ b/packages/rtfm/PKGBUILD
@@ -1,0 +1,43 @@
+# This file is part of BlackArch Linux ( http://blackarch.org ).
+# See COPYING for license details.
+
+pkgname='rtfm'
+pkgver=76.c281e6f
+pkgrel=1
+pkgdesc='A database of common, interesting or useful commands, in one handy referable form.'
+groups=('blackarch' 'blackarch-misc')
+arch=('any')
+url='https://github.com/leostat/rtfm'
+license=('custom:unknow')
+depends=('python2' 'python2-terminaltables')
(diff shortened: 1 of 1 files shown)


PedroSFreitas:
@noptrix I was reading the http://www.pathname.com/fhs/pub/fhs-2.3.html and something crossed my mind. Although I'm not a native English speaker and I could have understood something entirely wrong. Maybe you could shed some lights here.
In [/usr](http://www.pathname.com/fhs/pub/fhs-2.3.html#PURPOSE18) section you can see that 

> Large software packages must not use a direct subdirectory under the /usr hierarchy.

although [/usr/share](http://www.pathname.com/fhs/pub/fhs-2.3.html#PURPOSE26) section says:

> Any program or package which contains or requires data that doesn't need to be modified should store that data in /usr/share (or /usr/local/share, if installed locally). It is recommended that a subdirectory be used in /usr/share for this purpose

But [/opt](http://www.pathname.com/fhs/pub/fhs-2.3.html#PURPOSE14) section says:

> A package to be installed in /opt must locate its static files in a separate /opt/<package> or /opt/<provider> directory tree, where <package> is a name that describes the software package and <provider> is the provider's LANANA registered name.

So what do you think? Should I change this PR to install all the necessary files under /opt/${pkgname} (as [/opt](http://www.pathname.com/fhs/pub/fhs-2.3.html#PURPOSE14) section implies) or should we leave this way?

**Cheers**.

noptrix:
it's ok this way. thanks a lot :)

Issue #1759, opened 2017-08-31
[tool request] raven


PedroSFreitas:
https://github.com/0x09AL/raven

Note: I don't want to forget this and just to let others know that I'm already considering this tool.

PR #1760, opened 2017-09-02
[tool request] adding raven: #1759.

PedroSFreitas:
review of /opt or /usr/share according to my previous PR is necessary.


noptrix:
thanks a lot

PR #1785, opened 2017-10-04
Update and fix bro

p-l-:
Bro 2.5.1 is now the stable release (see
https://www.bro.org/download/).

Bro requires openssl-1.0 (see
<https://bro-tracker.atlassian.net/browse/BIT-1775>) and will not work
with Archlinux openssl package (this package is currently broken).

This commit also fixes the `build()` function (using cmake).


noptrix:
thank you

PR #1792, opened 2017-10-16
Belati Version Update 0.2.3-dev

aancw:
Hi @noptrix ,

I've updated Belati to version 0.2.3-dev. Add new package deps `perl-image-exiftool` and new python deps `pyexifinfo`. 

`pyexifinfo` not available at official repo, so i'm use pip instead, that's correct?

And the changelog is:
- Add Metadata/exif info for document
- Update Database Schema
- Auto Update system
- Update URL File validation checker
- Clean Output Result
- Update Banner
- Check git control status for update checker
- Add Web Server Only Mode
- Rework Code

I've implemented auto update system after 0.2.3-dev. So, Belati will doing  an automatic update from the Github repo when available.

Please correct me if my PKGBUILD is wrong :)

Thanks! :)


noptrix:
hey @aancw. thanks for this. belati is already upgraded. we use your git master branch, fyi.

PR #1799, opened 2017-10-23
Make Bupsuite able to be launched from GUI.

r-l8:
This pull request makes GUI launchers aware of Burpsuite.  As Burp is a GUI based application it seems reasonable that it should be able to be launched from a GUI. 

Also minor updates to the PKGBUILD that bring it in line with the Arch packaging style.

Keep up the great work!


noptrix:
thanks @r-l8 for this

r-l8:
I'm very happy that I was able to give something back :-) ! Sorry that I forgot to increment the pkgrel.

BTW, what is the reason to favour hardcoding the package name rather than using the pkgname variable? (I will avoid using it in the future)

PR #1800, opened 2017-10-25
Add package python-pyqtgraph

r-l8:
This pull request adds the package called `python-pyqtgraph`.

One of the optional dependencies for `binwalk` is `python-pyqtgraph`, which adds support for visualising entropy over a file. This package was not available in the core, extra, community, or blackarch repositories.


noptrix:
thank you

Issue #1825, opened 2017-12-06
[Tool request] slurp


vdun:
https://github.com/bbb31/slurp

PR #1833, opened 2017-12-19
[tool] adding slurp #1825

PedroSFreitas:
Adding `slurp` as requested on #1825.

Although I don't like cd'ing to the binary folder (usually `/usr/share/application`) and then call the binary, in this case I think it was necessary because it couldn't find the `permutation.json`. I didn't know what else I could do to avoid cd'ing, so I'm open to suggestion.


noptrix:
thx!

Issue #1839, opened 2018-01-01
[Tool removal request] pacaur


myrage2000:
Pacaur is now unmaintained :
As mentioned here : https://bbs.archlinux.org/viewtopic.php?pid=1755144#p1755144
The GitHub page is now ready-only : https://github.com/rmarquis/pacaur

Trizen seems to be a good aur helper.

PedroSFreitas:
@myrage2000 Thanks for bringing this up. I'll take a look at Trizen.

ReK2Fernandez:
yeah please add Trizen or something similar to Pacaur. tx <3

noptrix:
@myrage2000 Hi. Thanks for letting us know. Yep, trizen seems to be a good alternative although written in Perl...We are still reviewing few of the alternatives. Once done we will replace pacaur with $alternative and close this issue here.

ReK2Fernandez:
agghh perl.. noo so 2001 :P  ruby/python/go/rust <----  👍

noptrix:
thanks guys. pacaur has been removed and trizen is added, now.

PedroSFreitas:
Done.

PR #1844, opened 2018-01-14
trizen: requested on #1839

PedroSFreitas:
Adding trizen as a replacement for pacaur.
https://github.com/BlackArch/blackarch/issues/1839


noptrix:
thanks a lot

PR #1869, opened 2018-03-02
cisco-scanner: removing, not udpated since 2006.

PedroSFreitas:
To maintain the good quality of our tools, I believe removing this application
which was last updated back in 2006.
From the webpage: 'Last update: 08 March 2006 06:06:00'

Although feel free to ignore it if this tool is still relevant.


noptrix:
thanks for the reviews also for the other ones. let's keep them anyway. the clean-up of 'tool does not exist on $site' is not necessary. at least for now. let's ignore those

PR #1901, opened 2018-04-08
findsploit: fixing dep, groups and bumping pkgrel to force update.

PedroSFreitas:
- Upstream fixed the call to browser, so the `sed` for it isn't necessary anymore. (https://github.com/1N3/Findsploit/blob/master/findsploit#L14)

- I'm also adding `searchsploit` as dependency.

note: there were changes in the code (upstream) but the pkgver didn't change (git).


noptrix:
thx @PedroSFreitas

PR #1913, opened 2018-04-24
Fixed getsploit bug

adcar:
getsploit wouldn't launch due to not being in the second `getsploit` folder. This adds the second folder to the `/usr/bin/getsploit` file so it can be launched.


noptrix:
thanks a lot

PR #1936, opened 2018-06-03
syncrepo.sh: script for syncing, based on Arch.

PedroSFreitas:
- forked from Arch Linux syncing script, with changes to run out of the box.
- the packages will be stored under /srv/blackarch;
- the path to the packages should then be mount-bind'd to the appropriate place.


noptrix:
thanks mate

PR #1938, opened 2018-06-06
Updated kekeo: added pkgver() and updated URLs

osospeed:
https://github.com/gentilkiwi/kekeo/archive/... does not include binaries
https://github.com/gentilkiwi/kekeo/releases/download/.../kekeo.zip does


noptrix:
thank you @osospeed

PR #1942, opened 2018-06-09
BlackArch Tunning

4k4xs4pH1r3:
by 4k4xs4pH1r3

### DIFF
--- a/docs/BlackArch-tunning.md
+++ b/docs/BlackArch-tunning.md
@@ -114,6 +114,8 @@ pacman -S yaourt
 #
 pacman -S octopi
 #
+pacman -S pamac
+#
 pacman -S xorg xorg-server
 #
 pacman -S deepin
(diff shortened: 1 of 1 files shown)


PedroSFreitas:
First of all, thanks.
But this seems more like a personal customization than something that could be used by anyone. That said, I don't think it should be merged.

@noptrix what do you think?

4k4xs4pH1r3:
This pull request born from the issues related with network specifically getting up the interfaces; dhcp; wicd; d-bus / NetworkManager. The rest is value added

PedroSFreitas:
"Network getting up" is not an issue, is laziness of people that don't want to read the documentation.
About the value added, true. Good contribution. But as I said, very personal. For instance I wouldn't use any of that setup.
The idea is having the minimum possible related to configuration, so people can set their system the way they want; not the way we think they should.

4k4xs4pH1r3:
@PedroSFreitas This is the guide in english https://blackarch.org/blackarch-guide-en.pdf ?

Please share me the link where I can found the documentation for my case.

PedroSFreitas:
@4k4xs4pH1r3 thanks for taking the time to contribute, but we decided (as expressed above) that it should not be merged.

PR #1952, opened 2018-06-25
Make netcat implementation users choice for pentmenu

### LINKED ISSUE #1950: pentmenu requires gnu-netcat but recommends openbsd-netcat
Hey Blackarch Team,

I got a weird issue, and Im not sure where to point this out so I thought here might be a good start.
I installed Blackarch over an existing Manjaro Installation and that worked out fine.
Now I noticed that gnu-netcat is installed and I prefer openbsd-netcat. I tried to change it but couldnt deinstall gnu-nc because pacman says pentmenu depends on it.
I checked pentmenus git and it says "netcat (must support '-k' option, openbsd variant recommended)".
Deinstalling pentmenu, installing openbsd-netcat and then installing pentmenu from the AUR worked out fine but there is a newer version in the BlackArch Repository than in the AUR and now pacman wants to update pentmenu but says gnu-nc is required to do so.
Sounds like an error in your buildPKG as the package of the AUR is depending on openbsd-nc.
Thanks for your great work.

ikstream:
As mentioned in #1950 we have gnu-netcat as dependency, while openbsd-netcat could be used as well.
This resolves #1950 making it the choice of the user, which implementation to use.


noptrix:
@ikstream thank you

PR #1976, opened 2018-07-29
Autopsy 4.7.0

obilodeau:
a drastically different autopsy than from the 2.x era


noptrix:
thanks a lot @obilodeau

PR #1980, opened 2018-08-02
Updating depedencies list for Empire & Koadic

mrsmith0x00:
python2-xlrd is needed by Empire. (Empire don't run without this)
python2-tabulate is needed by Koadic. (Koadic don't run without this)


noptrix:
thank you

PR #1983, opened 2018-08-03
Fix gdbgui

ikstream:
- Add new package python-flask-compress as dependency
- update patch file
- remove tests directory to avoid file conflicts


noptrix:
thanks a lot @ikstream

Issue #1993, opened 2018-08-16
Burp 1.7.37


noraj:
`http://portswigger.net/DownloadUpdate.ashx?Product=Free` is giving 1.7.36, the last listed version is 1.7.36, `https://portswigger.net/burp/releases/download?product=community&version=1.7.36&type=jar` works but `https://portswigger.net/burp/releases/download?product=community&version=1.7.37&type=jar` doesn't exist yet for the free version.

I know 1.7.37 is in the release note but is not yet available for download (community edition). So switch back to `pkgver='1.7.36'`.

```
$ file /usr/share/burpsuite/burpsuite.jar
/usr/share/burpsuite/burpsuite.jar: HTML document, UTF-8 Unicode text, with very long lines, with CRLF line terminators
```

because `https://portswigger.net/burp/releases/download?product=community&version=1.7.37&type=jar` is redirecting to `https://portswigger.net/burp` so the PKGBUILD is actually downloading the HTML page...

see PR #1994

noptrix:
thanks a lot! we've fixed that in our scripts. re-built and re-added 1.7.36

PR #1994, opened 2018-08-16
1.7.37 doesn't exist yet

noraj:
see issue #1993


noptrix:
fixed as mentioned earlier. again thanks for this :)

PR #1996, opened 2018-08-17
fpdns: Fix install paths and dependencies

1c7718e7:
The fpdns package had a missing dependency (perl-net-dns) and install paths for perl modules were hardcoded for version 5.26.


noptrix:
thanks a lot

PR #2000, opened 2018-08-22
IVRE: switch to Python 3

p-l-:
Most dependencies are actually optional and are now treated as such. New optional dependencies have also been added.


noptrix:
@p-l- thank you!

PR #2001, opened 2018-08-22
New package: didier-stevens-suite

osospeed:


osospeed:
If it is totally windows tools then '/usr/share/windows' would be the place, but since it is mixed (windows and Linux) I think '/usr/share/' would fit here. No windows stuff should be put in '/usr/lib'.
Having windows tools in Arch is very helpful as security arsenal in one box. Also many tools work under wine.

FFY00:
Is the source code available? If so, the linux tools should be built from source and installed to `/usr/bin`, if not, they should do to `/opt/pkgname`. Windows binaries should always go to `/opt/pkgname` unless you are building them with mono or they are meant to be run with wine.

FFY00:
Just a final note, if the linux binaries are meant to be directly run, please symlink them to `/usr/bin` so that they are in `PATH` but not actually installed there. If not, the PKGBUILD seems fine to me, assuming the executable bits are correct in the upstream.

osospeed:
I prefer not to symlink and leave it as 'Use at your own risk'. Because:
- No source code available for the Linux binaries
- Not all *.py are tested under Linux

noptrix:
thanks for this, guys

PR #2020, opened 2018-09-06
Added new tool: onetwopunch

mrsmith0x00:
Use unicornscan to quickly scan all open ports, and then pass the open ports to nmap for detailed scans.

### DIFF
--- a/packages/onetwopunch/PKGBUILD
+++ b/packages/onetwopunch/PKGBUILD
@@ -1,0 +1,29 @@
+# This file is part of BlackArch Linux ( http://blackarch.org ).
+# See COPYING for license details.
+
+pkgname='onetwopunch'
+pkgver=17.d4ab4e8
+pkgrel=1
+pkgdesc='Use unicornscan to quickly scan all open ports, and then pass the open ports to nmap for detailed scans.'
+groups=('blackarch' 'blackarch-scanner')
+arch=('any')
+url='https://github.com/superkojiman/onetwopunch'
+license=('MIT')
+depends=('nmap' 'unicornscan')
+makedepends=('git')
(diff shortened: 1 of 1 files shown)


packages/onetwopunch/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+license=('MIT')
+depends=('nmap' 'unicornscan')
+makedepends=('git')
+source=('git+https://github.com/superkojiman/onetwopunch.git')
+sha1sums=('SKIP')
+
+pkgver() {
+  cd "$srcdir/onetwopunch"
+  
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
```
FFY00: The package provides a tag (release), it should be also in `$pkgver`. Use the following command instead.

```
git describe --long | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
```

packages/onetwopunch/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+}
+
+package() {
+  cd "$srcdir/onetwopunch"
+
+  install -Dm755 onetwopunch.sh "$pkgdir/usr/bin/onetwopunch"
+  install -Dm644 -t "$pkgdir/usr/share/doc/onetwopunch/" README.md
+  install -Dm644 LICENSE.md "$pkgdir/usr/share/licenses/onetwopunch/LICENSE.md"
```
FFY00: Please remove the `.md` from the installed file. The raw file doesn't need it. We should always install the plaintext version if we can.

https://raw.githubusercontent.com/superkojiman/onetwopunch/master/LICENSE.md

noptrix:
thanks guys. will do the changes

PR #2022, opened 2018-09-07
new tool 'winhex'

osospeed:


mrsmith0x00:
@osospeed On Arch Linux we have bless, ghex, hexedit and Hexinator eventually for hex editing. For recover MBR, files and other disk editing we have dd, testdisk, secure-delete and other. Why do we need WinHex? Or I don't know something? :)

osospeed:
I tested many other tools, but this one is special:
- Famous oldschool tool (but still updated)
- Additional templates files https://www.x-ways.net/winhex/templates/index.html
- Quite stable with big files
- Compare files
- Works under Windows and Linux (wine)
- All in one tool
- ... http://www.x-ways.net/winhex/allfeatures.html
- Always good to have alternatives :)

packages/winhex/PKGBUILD
```diff
@@ -0,0 +1,23 @@
+# This file is part of BlackArch Linux ( http://blackarch.org ).
+# See COPYING for license details.
+
+pkgname='winhex'
```
FFY00: No need for quotes. This might actually be a bad thing.

packages/winhex/PKGBUILD
```diff
@@ -0,0 +1,23 @@
...
+depends=()
+
+pkgver() {
+	curl 'https://www.x-ways.net/winhex/' | grep 'strong>WinHex ' | sed -r 's/.*WinHex ([0-9]+\.[0-9]+).*/\1/gi'
+}
+package() {
+	tmpFile='/tmp/winhex.zip'
+	targetDir="${pkgdir}/usr/share/windows/winhex/"
+	mkdir -p "${targetDir}"
+	curl -L --output "${tmpFile}" "https://www.x-ways.net/winhex.zip"
```
FFY00: Please don't. This is just wrong. Use the `sources()` instead.

packages/winhex/PKGBUILD
```diff
@@ -1,23 +1,24 @@
...
 }
+
 package() {
-	tmpFile='/tmp/winhex.zip'
-	targetDir="${pkgdir}/usr/share/windows/winhex/"
-	mkdir -p "${targetDir}"
-	curl -L --output "${tmpFile}" "https://www.x-ways.net/winhex.zip"
-	unzip -d "${targetDir}" "${tmpFile}"
-	rm -rf "${tmpFile}"
+  rm -rf './winhex.zip'
```
FFY00: Not needed.

packages/winhex/PKGBUILD
```diff
@@ -1,23 +1,24 @@
...
+  mkdir -p "${pkgdir}/usr/share/windows/winhex/"
+  cp ./* "${pkgdir}/usr/share/windows/winhex/"
```
FFY00: This should be installed to `/opt` instead since it doesn't seem to have the source available. Sorry for not mentioning it in the first review.

PR #2025, opened 2018-09-07
Added new tool: sigthief

mrsmith0x00:
Stealing Signatures and Making One Invalid Signature at a Time.

### DIFF
--- a/packages/sigthief/PKGBUILD
+++ b/packages/sigthief/PKGBUILD
@@ -1,0 +1,29 @@
+# This file is part of BlackArch Linux ( http://blackarch.org ).
+# See COPYING for license details.
+
+pkgname='sigthief'
+pkgver=14.211b4fe
+pkgrel=1
+pkgdesc='Stealing Signatures and Making One Invalid Signature at a Time.'
+groups=('blackarch' 'blackarch-exploitation' 'blackarch-windows')
+arch=('any')
+url='https://github.com/secretsquirrel/SigThief'
+license=('BSD')
+depends=('python')
+makedepends=('git')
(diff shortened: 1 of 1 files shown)


packages/sigthief/PKGBUILD
```diff
@@ -0,0 +1,29 @@
+# This file is part of BlackArch Linux ( http://blackarch.org ).
+# See COPYING for license details.
+
+pkgname='sigthief'
```
FFY00: No need for quotes. This might actually be a bad thing.

packages/sigthief/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+arch=('any')
+url='https://github.com/secretsquirrel/SigThief'
+license=('BSD')
+depends=('python')
+makedepends=('git')
+source=('git+https://github.com/secretsquirrel/SigThief.git')
+sha1sums=('SKIP')
+
+pkgver() {
+  cd "$srcdir/SigThief"
```
FFY00: No need to cd to `$srcdir`, we already start there.

Just:
```
cd SigThief
```

packages/sigthief/PKGBUILD
```diff
@@ -0,0 +1,29 @@
...
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
+
+package() {
+  cd "$srcdir/SigThief"
```
FFY00: No need to cd to `$srcdir`, we already start there.

Just:
```
cd SigThief
```

packages/sigthief/PKGBUILD
noptrix: i will change it

Issue #2009, opened 2018-09-01
Destiny of Windows tools


mrsmith0x00:
Hello everyone!

What about utilities which don't work under Wine and demand WinAPI low-level system calls for properly work? For example Intercepter-NG doesn't work (it crash at each tool) even after the official wine_pcap_dlls installation. OllyDbg can't work with some Win32 apps (on Windows OllyDbg work perfectly). x64dbg at all can't start with Wine. Nowadays Wine can't normally "pass" Windows low-level system calls to GNU/Linux. We have a lot of pretty and more useful utilities for GNU/Linux. What about to delete them and similar?

P.S. I can make list of them.

Tazmain:
Is this not more a bug with wine currently? But I think it might be worth looking into.

mrsmith0x00:
@Tazmain It is possible to call also a bug. But support of some libraries leaves much to be desired (undocumented opportunities, etc.). Wine developers especially also don't try to correct it (though whether it's possible in general?).

mrsmith0x00:
Who else what thinks about it?

FFY00:
I don't think they necessarily need to have platinum on AppDB. Just try them out and if the main functions work, let them be.

ikstream:
I think silver status would be good enough, as silver status apps should work correct with just some minor issues.

mrsmith0x00:
Ok, so soon, as I will be able, I will begin to check each utility.

FFY00:
You could keep this open and report when you finish.

mrsmith0x00:
What about idea to store the all Windows programs in some directory (for example the same /usr/share/windows/), instead of delete or to try to force them to work? For example, present a situation: access to the machine (Windows) is got, but it is necessary to scan network from within (or do any other thing). Then we can just copy the necessary file and work! I think we should only delete any scripts for run unsupported tools in GNU/Linux and make it an perfect archive of auxiliary tools. What do you think about it?

mrsmith0x00:
For example:
OllyDbg works normal - we leave it as program.
x64dbg crash on start - we leave it only as a zip archive in /usr/share/windows/ (or another directory) and no more.

FFY00:
I don't think there's any need to leave there a program that doesn't work.

mrsmith0x00:
Even if works from under Windows?

FFY00:
Why do you need them? Is it that common to run windows VMs in pentesting setups?

mrsmith0x00:
[...]

PR #2029, opened 2018-09-08
Clean up windows utilities

mrsmith0x00:
All tools tested on Wine 3.15 with all fonts, `comctl32ocx dotnet40 dotnet45 l3codecx mfc40 mfc42 vb6run vcrun{2003,2005,2008,2010,2012,2013,2015,2017} vcrun6`, installed by winetricks. Also with lib32-mpg123 package.

Some proofs of errors:


mrsmith0x00:
It's a lot of idle utilities - [ntsecurity.nu](http://ntsecurity.nu/toolbox/). On native Windows 7 they doesn't work, because these utilities mostly made for Windows NT 5.0 and earlier. Nowadays it's infrequently possible to see Windows 2000 on PC (and it's very good ;)
I removed only doesn't work GUI and more than half ntsecurity tools (and absolutely useless, as netbus).

noptrix:
thanks @mrsmith0x00 :)

osospeed:
As I understood form the discussion (https://github.com/BlackArch/blackarch/issues/2009#issuecomment-419470066), tools that work under Windows but not Linux should be moved to `/opt` !

mrsmith0x00:
@osospeed sure, as I will deal with tools check, I will create PR with transfer in a `/opt`.

PR #2050, opened 2018-09-17
Add tld dependency to photon package

y0no:


noptrix:
@y0no thank you. like with the python{2}-tld packages: remove the quotes, etc. i will do that myself now.

PR #2056, opened 2018-09-22
[tool] Adding bettercap

goabonga:


noptrix:
hi @goabonga thanks a lot for this. though, arch community has bettercap already. so there is no need to re-create it from our side. it is maintained by ArchLinux. We are mirroring the package also.

goabonga:
Hooo, sorry, I had not seen.
(thanks to Christian Rebischke for that)

PR #2102, opened 2018-11-08
add missing dep

noraj:
fix a dependency issue

  File "mr.sip.py", line 37, in <module>                                                                                                                                                                          
    import random,string,ipaddress,netifaces,os,socket,logging                                                                                                                                                    
ImportError: No module named netifaces


noptrix:
thanks @noraj

PR #2137, opened 2018-12-06
create python3 pdfkit package

noraj:


noraj:
@ikstream right, because I based the PKGBUILD on the python2 version, so the change is also needed here https://github.com/BlackArch/blackarch/pull/2144

noptrix:
@noraj well ok, let me fix these things. you can then see the changes and next time try to keep the same style etc.

noptrix:
@noraj well ok, let me fix these things. you can then see the changes and next time try to keep the same style etc.

noptrix:
@noraj jfyi: https://github.com/BlackArch/blackarch/commit/276553eb4bc5903b39828c3d872312e978028d92

noraj:
@noptrix Ok, my bad.

PR #2138, opened 2018-12-06
[WIP] create python3 androguard

noraj:


noraj:
I'm not sure this is respecting all dependencies https://github.com/androguard/androguard/blob/master/setup.py

ikstream:
The requirements should be added as dependencies in the PKGBUILD file.
I am not sure how to handle the additional requirements. but at least they should be optional dependencies.
Have you tried to build and run it in a clean blackarch environment?

noraj:
@ikstream https://github.com/androguard/androguard/blob/a18256203d7af751c8862b04b15b15c23225b54f/setup.py#L17-L37
This is targeting python3 on archlinux which is currently `3.7.1` but all the dependency so are only for python < 3.4, so in our case the only mandatory dependency is `ipython>=5.0.0`.

noraj:
I added `iptyhon` dependency, optdepends and checkdepends. Only `nose-timer` (test) and `sphinxcontrib-programoutput` (optinnal) are missing.

packages/python-androguard/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+# See COPYING for license details.
+
+pkgname=python-androguard
+pkgver=3.2.1
+pkgrel=1
+pkgdesc='Androguard is a full python tool to play with Android files.'
+arch=('any')
+url='https://pypi.org/project/androguard'
+license=('MIT')
+depends=('python'
```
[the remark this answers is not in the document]
ikstream: They are needed. The `install_requires` are not in the python 3.4 indention block and the different scrip[ts wont work without them

packages/python-androguard/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+         'iptyhon')
+makedepends=('python-setuptools')
+optdepends=('python-pyperclip'
+            'python-pyqt5'
+            'python-magic'
+            'python-filemagic'
+            'python-sphinx'
+            'python-sphinx_rtd_theme'
+            'python-pydot')
+checkdepends('python-mock'
```
ikstream:
```suggestion
checkdepends=('python-mock' 'python-nose' 'python-codecov' 'python-coverage')
```

ikstream: Another typo in here.

packages/python-androguard/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+pkgver=3.2.1
+pkgrel=1
+pkgdesc='Androguard is a full python tool to play with Android files.'
+arch=('any')
+url='https://pypi.org/project/androguard'
+license=('MIT')
+depends=('python' 
+         'iptyhon')
+makedepends=('python-setuptools')
+optdepends=('python-pyperclip'
```
ikstream: Add some info, what the packages are used for
https://wiki.archlinux.org/index.php/PKGBUILD#optdepends

packages/python-androguard/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+
+  python setup.py build
+}
+
+package() {
+  cd "androguard-$pkgver"
+
+  python setup.py install --prefix=/usr --root="$pkgdir" --optimize=1 \
+    --skip-build
+}
```
ikstream:
```suggestion
}

```

packages/python-androguard/PKGBUILD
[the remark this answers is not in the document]
noraj: @ikstream You are right, I misread it, only the print is in the if 3.4 scope, that is the problem with language like python that use indentation for blocks. 
And for the test I apologize but i'm building this PKG only as a dependency of another so I don't really know how to test it.

packages/python-androguard/PKGBUILD
noraj: I will

noraj:
@ikstream I think it is all fixed, I can test right now I'm on Winshit 10.

packages/python-androguard/PKGBUILD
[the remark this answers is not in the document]
ikstream: Was very likely me, as i was copying from file into Github.
How to test depends on the package that is added. androguard for example provides some
command line tools you can call.
If it's a python library, just try to import it in python repl and probably do a basic call, this will show most problems with dependencies.

ikstream:
I will try to test it later this day

Edu4rdSHL:
Closing it because the actual androguard version is already in python3. See https://github.com/BlackArch/blackarch/commit/42a3e0194aeb8ebe2da312be8e412125dea61f18

ikstream:
@Edu4rdSHL Most of the androguard stuff is not running atm, as depencies are missing. I will test this pr first, but would stick with it, as it schould be running. What's you opinion on this?

ikstream:
Just tested it and works for me.

Edu4rdSHL:
@ikstream since that the androguard package was updated yesterday to use python3 instead of python2 (see https://github.com/BlackArch/blackarch/commit/42a3e0194aeb8ebe2da312be8e412125dea61f18) I don't see reason to merge it commit into master and also i dont see reason to have a python2 package of a project that is using python3.

noraj:
@Edu4rdSHL I'm fine with that.

Edu4rdSHL:
Thank you for all the work, @noraj.

ikstream:
@Edu4rdSHL have you checked the androguards tools after the update? With the PKGBUILD file of @noraj  they work for me, with the currently available package they don't.

I am with you, that keeping two package makes no sense

Edu4rdSHL:
@ikstream yes, it's working as expected, you can confirm that.

ikstream:
Did you run it in a clean environment? When i try to start the command line tools, at least `asn1crypto` is missing. probably more are missing. I recommend using the PKGBUILD file, as it also points out optional dependencies.

Edu4rdSHL:
@ikstream thank you, all should be fixed now.

PR #2198, opened 2019-01-22
eyewitness: add missing dependencies

mzet-:


noptrix:
@mzet- thanks a lot again.

PR #2205, opened 2019-01-25
add missing deps

noraj:
according to https://github.com/NORMA-Inc/AtEar#dependency-programs

but ` python2-sqlite` is still missing for example


noraj:
@noptrix sorry I missed that

PR #2221, opened 2019-02-07
rename 'python2-libbde-python' -> 'python2-libbde'

osospeed:


mrsmith0x00:
@osospeed I fully agree with you, that name without `python` pretty, but this destroy auto-update for this package. I'm sorry, but we can't merge commit.

osospeed:
Sorry, I did not understand exactly what do you mean by 'destroy auto-update'. Can you elaborate more ?

mrsmith0x00:
@osospeed script `version-checker.py` get [python-something] names from `packages/` directory and then connect to PyPI JSON API to check for package updates. If you change package name (both in `pkgname` and in `packages/` directory, script will not find package in PyPI repo.

osospeed:
OK. So if we build it from git instead of pypi, would it then be renamed as 'python2-libbde' ?

mrsmith0x00:
@osospeed btw, why build from git, instead of use from PyPI? It's a bit faster than do `git clone` on each package.

PR #2223, opened 2019-02-07
new tool: python2-libesedb

osospeed:


mrsmith0x00:
@osospeed and please, use package from PyPI, instead of git!

PR #2235, opened 2019-02-11
New tool: python2-libsmraw-python

osospeed:


noptrix:
@osospeed JFYI: i will remove the groups section since these are libs and don't provide any programs. so you will need to explicitly say ```pacman -S python-<whatever>``` rather than ```pacman -S blackarch --needed```

PR #2239, opened 2019-02-12
Rename: python2-libesedb -> python2-libesedb-python

osospeed:
To match PyPI name and having automatic updates.


osospeed:
Old fix ;)

PR #2241, opened 2019-02-12
New tool: python2-libfvde-python

osospeed:


FFY00:
Not all fine, `gcc` shouldn't be in the `makedepends` as it is part of the `base-devel` group.

noptrix:
> Not all fine, `gcc` shouldn't be in the `makedepends` as it is part of the `base-devel` group.

Thanks for this. Will fix it in the previous ones. @osospeed can you fix that for this PR, please?

osospeed:
Fixed in: https://github.com/BlackArch/blackarch/pull/2242

PR #2252, opened 2019-02-23
dvcs-ripper: add missing scripts and deps

noraj:
package the whole tool https://github.com/kost/dvcs-ripper


noraj:
@noptrix I'm glad to be involved in blackarch and I hate hit and runners so I can't afford to be one :)

PR #2253, opened 2019-02-24
add aws-extender-cli

noraj:
I tested it works perfectly. `aws-extender-cli -b xxxxxx -s S3`

### DIFF
--- a/packages/aws-extender-cli/PKGBUILD
+++ b/packages/aws-extender-cli/PKGBUILD
@@ -22,17 +22,17 @@ pkgver() {
 
 package() {
   cd $pkgname
-  
+
   install -dm 755 "$pkgdir/usr/bin"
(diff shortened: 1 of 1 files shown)


noptrix:
> thank you! almost fine. i will merge in this time and make some changes. you can see afterwards and next time you can try to make the same :)

@noraj https://github.com/BlackArch/blackarch/commit/cfce26d50889f4e5bb81c11dedcf4794794ee8f9

noraj:
@noptrix Please explain me. 

1.What is the difference between `blackarch-scan` and `blackarch-scanner`?
2. Why `blackarch-config` is not appropriate, because the tool is finding and reviewing config files to find some information?
3. Why `source=('git+https://github.com/VirtueSecurity/aws-extender-cli.git')` instead of `source=("git+$url")` if it is the same URL?

PR #2269, opened 2019-03-08
xortool: upgrade + dependency fix

noraj:
solve this (see https://github.com/hellman/xortool/blob/master/setup.py)

```
xortool -h            
  File "/usr/bin/xortool", line 39, in <module>
    from xortool.args import parse_parameters, ArgError
  File "/usr/lib/python2.7/site-packages/xortool/args.py", line 4, in <module>
    from docopt import docopt
ImportError: No module named docopt
```

And is now compatible with python 3

### DIFF
--- a/packages/xortool/PKGBUILD
+++ b/packages/xortool/PKGBUILD
@@ -9,7 +9,7 @@ pkgdesc='A tool to analyze multi-byte xor cipher.'
 arch=('any')
 url="https://github.com/hellman/xortool/"
 license=('GPL')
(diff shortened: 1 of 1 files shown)


noraj:
I tested an it works well.

noraj:
@Edu4rdSHL Ok, but can I know why ?

Edu4rdSHL:
@noraj because we are using the  `python-docopt` [community] package, it's in 0.6.2 and ArchLinux doesn't have archives in their repos that can allow users to install another minor version than it. Specifying the explicit package version is unnecessary.

PR #2270, opened 2019-03-09
RsaCtfTool: upgrade

noraj:
for this error

```
  File "RsaCtfTool.py", line 15, in <module>
    from Crypto.PublicKey import RSA
ImportError: No module named Crypto.PublicKey
```

`PyCrypto` is missing but not available in arch or blackarch repos. bur is available in AUR https://aur.archlinux.org/packages/python-pycrypto/

requirements.txt: https://github.com/Ganapati/RsaCtfTool/blob/master/requirements.txt

https://github.com/Ganapati/RsaCtfTool/blob/master/RsaCtfTool.py#L1 = `#!/usr/bin/env python3`

### DIFF
--- a/packages/rsactftool/PKGBUILD
+++ b/packages/rsactftool/PKGBUILD
@@ -9,8 +9,8 @@ groups=('blackarch' 'blackarch-crypto' 'blackarch-cracker')
 arch=('any')
 url='https://github.com/Ganapati/RsaCtfTool'
 license=('custom:unknown')
-depends=('python2' 'python2-sympy' 'python2-argparse' 'python2-gmpy'
-         'python2-libnum')
+depends=('python' 'python-sympy' 'python-gmpy'
+         'python-requests' 'python-pycrypto')
 makedepends=('git')
 source=('git+https://github.com/Ganapati/RsaCtfTool.git')
 sha512sums=('SKIP')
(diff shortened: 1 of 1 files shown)


noraj:
tested: totally works (with python-pycrypto installed from AUR before) so we just need to add it in blackarch too?

packages/rsactftool/PKGBUILD
```diff
@@ -9,8 +9,8 @@ groups=('blackarch' 'blackarch-crypto' 'blackarch-cracker')
 arch=('any')
 url='https://github.com/Ganapati/RsaCtfTool'
 license=('custom:unknown')
-depends=('python2' 'python2-sympy' 'python2-argparse' 'python2-gmpy'
-         'python2-libnum')
+depends=('python' 'python-sympy' 'python-gmpy2'
+         'python-requests' 'python-pycrypto')
```
Edu4rdSHL: Change `python-pycrypto` for `python-pycryptodome`, it provides the Crypto module.

noraj: I didn't know pycryptodome was providing a different implementation of this class.

noraj:
Re-tested: works with pycryptodome

Edu4rdSHL:
@noraj thank you. Just other recommendation, in the package() function use `cp -a --no-preserve=ownership` instead of simply `cp -a *`, I will change it when merge the PR.

noraj:
@Edu4rdSHL Ok, nice to know. But it was already there, I didn't changed that part.

PR #2271, opened 2019-03-11
tools added

AaronS28:


packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+pkgname=ntopng
+pkgver=3.8
+pkgrel=1
+pkgdesc='Web-based Traffic and Security Network Traffic Monitoring.'
+arch=('any')
+groups=('blackarch')
+url='https://www.ntop.org/'
+license=('GPL-3.0')
+depends=('gcc' 'libmaxminddb' 'nDPI')
```
Edu4rdSHL: nDPI is not in the repos, you need to package it one different PKGBUILD.

packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+makedepends=('nDPI')
+source=("https://github.com/ntop/ntopng/archive/3.8-stable.zip")
```
Edu4rdSHL: Use $pkgname and $pkgver variables even if is possible.

packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+sha512sums=('SKIP')
```
Edu4rdSHL: If you're downloading a non git source, sha512sums doesn't can be skipped.

packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+prepare() {
+   cd "$pkgname-$pkgver-stable"
+   ./autogen.sh
+}
+
+build() {
+  ./install_nDPI.sh
```
Edu4rdSHL: Why are you doing that? You can't just run any `sh` script in the PKGBUILD, as I said, if you need ndpi for the package, then you need to package it using lowercase names.

packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+pkgname=ntopng
+pkgver=3.8
+pkgrel=1
+pkgdesc='Web-based Traffic and Security Network Traffic Monitoring.'
+arch=('any')
+groups=('blackarch')
+url='https://www.ntop.org/'
+license=('GPL-3.0')
+depends=('gcc' 'libmaxminddb' 'nDPI')
+makedepends=('nDPI')
```
Edu4rdSHL: Why a make-dependency is in the dependency array too? It can't be the same.

packages/ntopng/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+build() {
+  ./install_nDPI.sh
+  cd "$pkgname-$pkgver-stable"
+  ./configure
+}
+
+package() {
+  cd "$pkgname-$pkgver-stable"
+  sudo make
```
Edu4rdSHL: You never need to use sudo in PKGBUILDs.

packages/ntopng/install_nDPI.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
Edu4rdSHL: It file need to be deleted, if you need `ndpi`, package it.

Edu4rdSHL:
Closing because not answer was provided by trhe requester.

PR #2279, opened 2019-03-17
Add python-miasm-git and python2-miasm-git

p-l-:
Miasm (https://miasm.re / https://github.com/cea-sec/miasm) is a reverse engineering framework.

No release has yet been made since Miasm is "packageable" (i.e., depends only on resources that are already in Arch Linux), so for now we can only have the -git packages. I'll create another PR for the next stable release.

As a BlackArch & Miasm user, I can maintain these packages.


p-l-:
Thanks @noptrix! I will.

PR #2282, opened 2019-03-19
Updated Kickthemout PKGBUILD to use python3

Baitinq:
Updated Kickthemout PKGBUILD to use python3, as python2 made it so there was an error


Edu4rdSHL:
Hello, you need to modify the dependencies(), the pkgver() and test the package first, not only the executing order. Please take a look to https://github.com/BlackArch/blackarch/commit/4d8029445ed4a97a1e15166b1fe05efa2af332aa for future references. Thanks and we'r waiting your next PRs.

PR #2298, opened 2019-04-05
theharvester: Correct dependencies

flacks:
python-yaml is needed, not python-pyaml. python-retrying is also needed.


noptrix:
thank you!

flacks:
you're welcome. I forgot to bump pkgrel, though.

PR #2300, opened 2019-04-06
ghidra: Package fixes

flacks:
Ghidra contains many 64-bit ELF executables, including one critical to
Ghidra's proper functioning, therefore should only be packaged as an
x86_64 package.

Fix typo for java-environment in dependencies.

Remove optional dependencies - Python is not needed as Ghidra uses
Jython; however, it is not an optional dependency either as it is
already bundled as a standalone environment within Ghidra; finally, PAM
is not supported by Ghidra Server.


flacks:
Since Ghidra source is now available, I've submitted ghidra-git to the AUR. Shouldn't this package build from source?

https://aur.archlinux.org/packages/ghidra-git

osospeed:
@flacks Sure, please go ahead and add it here!

PR #2303, opened 2019-04-08
dvcs-ripper: dependency fix

noraj:


noraj:
@noptrix My bad for pkgrel, I did it locally but did not pushed it.

PR #2331, opened 2019-05-07
Webshells: minor changes

noraj:
+ remove readme from /usr/share/webshells/
+ add a post-install info about where webshells were copied


noptrix:
thanks a lot @noraj

Issue #2332, opened 2019-05-08
creddump: multiple problems


noraj:
Reference PKGBUILD: https://github.com/BlackArch/blackarch/blob/master/packages/lfimap/PKGBUILD

1. The source `https://code.google.com/p/creddump/` is dead
2. Automatic export `https://github.com/moyix/creddump` has no release
3. The way it is packaged right now is bugged, it will lead to the following error, because `install -Dm 755 -t "$pkgdir/usr/share/windows/$pkgname" cachedump.py lsadump.py pwdump.py` is used and the `framework` lib folder is not copied.
```
Traceback (most recent call last):
  File "/usr/share/windows/creddump/pwdump.py", line 25, in <module>
    from framework.win32.hashdump import dump_file_hashes
ImportError: No module named framework.win32.hashdump
```
4. Scripts are copied into `/usr/share/windows/$pkgname` so there is not tool available in the PATH.
5. The original 0.3 package requires `python-crypto` (aka pycrypto) that is not satisfied and need to be moved to `pycryptodome` anyway.

So I forked `https://github.com/moyix/creddump` to `https://github.com/noraj/creddump`, made the `sed -i 's/env python/env python2/' *.py` change and created a 0.3 release. And made a better PKGBUILD based on that.

I'll link the PR.

Edu4rdSHL:
Fixed via https://github.com/BlackArch/blackarch/commit/b85c0d1421a8cd018b72cbe47bff9c4222dd06ba

noraj:
@Edu4rdSHL 

Why not doing 

`source=("$pkgname::git+$url.git")`

rather than

`source=("$pkgname::git+https://github.com/moyix/$pkgname.git")`

?

Edu4rdSHL:
See the PKGBUILD-generic-git template in https://github.com/BlackArch/blackarch-pkgbuilds

noraj:
@Edu4rdSHL Yes but why? For readability ?

Edu4rdSHL:
Yeah, that's one of the most important things for us: make easy to understand for new users that want to contribute.

noraj:
@Edu4rdSHL Good reason so ;)

PR #2333, opened 2019-05-08
creddump rework

noraj:
#2332

### DIFF
--- a/packages/creddump/PKGBUILD
+++ b/packages/creddump/PKGBUILD
@@ -3,27 +3,44 @@
 
 pkgname=creddump
 pkgver=0.3
-pkgrel=7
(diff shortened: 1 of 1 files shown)


noraj:
PS: I fully tested this change with `blackarch-devtools`

noraj:
Should I add a `provides()` for the 3 scripts?

Edu4rdSHL:
We doesn't want to switch from upstream repos if there isn't *sufficient  reasons* and definitively there aren't. I will take some of your work and update the PKGBUILD.

noraj:
b85c0d1421a8cd018b72cbe47bff9c4222dd06ba

PR #2334, opened 2019-05-08
weevely: minor fixes to the packaging

noraj:
- `install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"` is useless, see https://wiki.archlinux.org/index.php/PKGBUILD#license
- install the man page

### DIFF
--- a/packages/weevely/PKGBUILD
+++ b/packages/weevely/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=weevely
 pkgver=829.8036a61
-pkgrel=1
+pkgrel=2
 groups=('blackarch' 'blackarch-webapp' 'blackarch-backdoor')
 pkgdesc='Weaponized web shell.'
 arch=('any')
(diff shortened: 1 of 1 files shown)


noraj:
It has been tested with `blackarch-devtools`.
Is it required to call mandb after adding a page? I went to the chroot and did `man 1 weevely` without having to call mandb. Also I see that other packages don't call mandb either: https://github.com/BlackArch/blackarch/blob/7b6007ab751cb9c1faeb6606caa70b11923eec5f/packages/nipper/PKGBUILD

Edu4rdSHL:
> *  `install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"` is useless, see https://wiki.archlinux.org/index.php/PKGBUILD#license

See #2 here:

2. Install the license in: /usr/share/licenses/pkgname/, e.g. /usr/share/licenses/foobar/LICENSE. One good way to do this is by using:

`install -Dm644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"` that's what we're doing.

> * install the man page

There isn't needed to use -dm before -Dm, you can simply do:

`install -Dm 644 weevely.1 "$pkgdir/usr/share/man/man1/weevely.1"`

Closing for that reasons, I will add the manpage.

noraj:
@Edu4rdSHL 

> Install the license in: /usr/share/licenses/pkgname/, e.g. /usr/share/licenses/foobar/LICENSE. One good way to do this is by using:

But this is after:

> The licenses package (part of the base group) contains many commonly used licenses, which are installed under /usr/share/licenses/common/. If a package is licensed under one of these licenses, the value should be set to the directory name, e.g. license=('GPL'). If the appropriate license is not included, several things must be done: 

So either:

1. It has a common license listed under `/usr/share/licenses/common/` and you must use`license=('GPL')`
2. OR **If the appropriate license is not included, several things must be done:**
  1. `license=('custom)` or `license=('custom:name')`
  2. `install -Dm644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"`

But if the step you mentioned **MUST** be done **ONLY** for custom licenses that are not part of the `licenses`. Please read https://wiki.archlinux.org/index.php/PKGBUILD#license carefully.

I'm sorry to insist.

PR #2354, opened 2019-05-26
SecList+domLink

vladimircicovic:
https://github.com/danielmiessler/SecLists
https://github.com/vysecurity/DomLink


noptrix:
@vladimircicovic thanks but i really don't understand what you did here... also, 'domlink' has been added now, although a really lame tool and still using python2 (we had to patch the code because of broken indentations...). seclists is too big to add. it is also provided via AUR  (seclists-git)

PR #2363, opened 2019-06-14
searchsploit: add conflicting package

noraj:
conflicts with https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/exploitdb

@noptrix Also should we remove this package as arch exploitdb is already providing it?

Where is the BA `exploit-db` wrapper is ?


noptrix:
@noraj yes true. let's remove searchsploit and mirror exploitdb. i will close this issue then.

PR #2364, opened 2019-06-14
cmsmap: missing dependency

noraj:
need `exploitdb`

`/usr/share/exploitdb/` is already the good value for `edbpath` in `cmsmap.conf`


noptrix:
@noraj thanks a lot

PR #2380, opened 2019-06-20
fix python-macholib

s7x:
Fixing the binaries overloap when python-macholib and python2-macholib gets installed.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-macholib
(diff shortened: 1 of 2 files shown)


noptrix:
@s7x ok @ikstream is right here. normally we should use conflicts argument. but when for example tool_x depend on python2-macholib and tool_y on python-macholib we can safely rename the py2 bins here as you did. for now lt's use the conflicts argument. i will change that

s7x:
@noptrix and @ikstream Ok guys thanks for the heads up, I didn't know about this one ;)

PR #2401, opened 2019-07-07
Dirble: update license and version numbering

sciguy16:
I've updated the pkgbuild for dirble to match its versioning scheme and license

### DIFF
--- a/packages/dirble/PKGBUILD
+++ b/packages/dirble/PKGBUILD
@@ -2,40 +2,34 @@
 # See COPYING for license details.
 
 pkgname=dirble
-pkgver=1.4.0
+pkgver=1.4.2
 pkgrel=1
+epoch=1
 pkgdesc='Fast directory scanning and scraping tool.'
 groups=('blackarch' 'blackarch-webapp' 'blackarch-scanner')
 arch=('x86_64' 'armv6h' 'armv7h' 'aarch64')
 url='https://github.com/nccgroup/dirble'
 license=('GPL3')
-makedepends=('git' 'cargo' 'rust')
-source=('git+https://github.com/nccgroup/dirble.git')
+makedepends=('cargo')
+source=("$pkgname::https://github.com/nccgroup/dirble/archive/v$pkgver.tar.gz")
 sha512sums=('SKIP')
 
-pkgver() {
-  cd $pkgname
-
-  awk -F'"' '/^version/ {print $2}' Cargo.toml
-}
 
 build() {
-  cd $pkgname
+  cd "$pkgname-$pkgver"
 
-  make
+  cargo build --release
 }
 
 package() {
-  cd $pkgname
-
-  install -dm 755 "$pkgdir/usr/share/$pkgname"
+  cd "$pkgname-$pkgver"
 
   install -Dm 755 "target/release/$pkgname" "$pkgdir/usr/bin/$pkgname"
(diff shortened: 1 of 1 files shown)


Edu4rdSHL:
I don't understand why we're copying files when we can juts run `cargo build --release` and then use `install -Dm 755 target/release/$pkgname "$pkgdir/usr/bin/$pkgname"`

Any specific reason?

Also, if you want to use a minor `pkgver` number, you need to add `epoch`, see https://wiki.archlinux.org/index.php/PKGBUILD#epoch

packages/dirble/PKGBUILD
```diff
@@ -2,21 +2,21 @@
 # See COPYING for license details.
 
 pkgname=dirble
-pkgver=186.c1b7237
+pkgver=1.4.0
```
Edu4rdSHL: Add epoch.

packages/dirble/PKGBUILD
```diff
@@ -2,21 +2,21 @@
...
 groups=('blackarch' 'blackarch-webapp' 'blackarch-scanner')
 arch=('x86_64' 'armv6h' 'armv7h' 'aarch64')
 url='https://github.com/nccgroup/dirble'
-license=('MIT')
+license=('GPL3')
 makedepends=('git' 'cargo' 'rust')
 source=('git+https://github.com/nccgroup/dirble.git')
 sha512sums=('SKIP')
 
 pkgver() {
```
Edu4rdSHL: Remove the whole pkgver() function if you're going to use versioned model.

packages/dirble/PKGBUILD
```diff
@@ -2,21 +2,21 @@
...
-pkgver=186.c1b7237
+pkgver=1.4.0
 pkgrel=1
 pkgdesc='Fast directory scanning and scraping tool.'
 groups=('blackarch' 'blackarch-webapp' 'blackarch-scanner')
 arch=('x86_64' 'armv6h' 'armv7h' 'aarch64')
 url='https://github.com/nccgroup/dirble'
-license=('MIT')
+license=('GPL3')
 makedepends=('git' 'cargo' 'rust')
```
Edu4rdSHL: Remove git if you're going to use versioned model.

packages/dirble/PKGBUILD
```diff
@@ -2,21 +2,21 @@
...
 source=('git+https://github.com/nccgroup/dirble.git')
```
Edu4rdSHL: Change the source pointing to the release version (like `$pkgname::https://github.com/nccgroup/dirble/archive/v$pkgver.tar.gz`)

ikstream:
I think it was on purpose, that we use with git `pkgver` instead of releases

packages/dirble/PKGBUILD
```diff
@@ -2,21 +2,21 @@
...
 sha512sums=('SKIP')
 
 pkgver() {
   cd $pkgname
 
-  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+  awk -F'"' '/^version/ {print $2}' Cargo.toml
 }
 
 build() {
```
Edu4rdSHL: Use `cargo build --release` instead of make.

sciguy16:
Thanks for the pointers, I've updated it to fully commit to versioned releases. It only needs cargo as a make dependency, because cargo implies rust. And namcap pointed out that libcurl is a runtime dependency, so I've added that too.

PR #2434, opened 2019-08-12
tor-browser-en: zenity dependency.

PedroSFreitas:
zenity is not 'optdepends' anymore, it is mandatory.


Edu4rdSHL:
Hello, thanks @PedroSFreitas. Is there any doc where tor-browser explicitly requires zenity as dependency? It works fine for me as it's actually.

PedroSFreitas:
> Hello, thanks @PedroSFreitas. Is there any doc where tor-browser explicitly requires zenity as dependency? It works fine for me as it's actually.

My tests were performed on a container (systemd-nspawn) using a vanilla Arch Linux installation with BlackArch repository (as all my tests when submitting a PR). Without Zenity the Tor Browser wouldn't start giving the following error message:

image

If it is working on your system now it could be because you already have zenity installed as a dependency for another package.

Edu4rdSHL:
Thanks. Please bump the `pkgrel` in order to merge it.

Edu4rdSHL:
Thank you, @PedroSFreitas.

PR #2453, opened 2019-09-01
Fix major bug

staz0t:
Fixed major logic bug in the latest release

### DIFF
--- a/packages/hashcatch/PKGBUILD
+++ b/packages/hashcatch/PKGBUILD
@@ -4,8 +4,8 @@
 # Submitter: noptrix
 
 pkgname=hashcatch
-pkgver=48.478e93c
-pkgrel=1
+pkgver=49.86efef5
+pkgrel=2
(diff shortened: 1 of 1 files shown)


staz0t:
Hey,
But i have released a separate package for this in my GitHub since it was a major fix. You can see 2 releases in hashcatch's releases. So, should I still keep $pkgrel as 1?

noptrix:
> Hey,
> But i have released a separate package for this in my GitHub since it was a major fix. You can see 2 releases in hashcatch's releases. So, should I still keep $pkgrel as 1?

yes, please. this hasn't anything to do with your releases. it has to do with blackarch.db

staz0t:
Done, I've made the changes. Just out of curiosity, when does one increase the pkgrel?

noptrix:
> Done, I've made the changes. Just out of curiosity, when does one increase the pkgrel?

when you for example change something in the pkgbuild only to fix the install but the version (pkgver) didn't change at all.

staz0t:
Got it, thanks!

Issue #2454, opened 2019-09-03
network pivoting tools to add


noraj:
(network pivoting) Tools I should add in the next days : 

- [x]  https://github.com/L-codes/Neo-reGeorg #2460
- [ ]  https://github.com/r3vn/punk.py

PR #2460, opened 2019-09-05
add Neo-reGeorg + minor fixes

noraj:
- add Neo-reGeorg #2454
-  minor PKGBUILD impovments
    - chisel: add group
    - lyricpass: remove useless makedependency since python setup.py build
    is not used
    - regeorg: fix license name & wrong groups


noraj:
For travis CI `packages/chisel/PKGBUILD:40:1: E005 Unquoted $srcdir or $pkgdir` all our PKGBUILDs are using `cd $srcdir` and not `cd "$srcdir"` is that a problem or should we remove this CI rule?

PR #2467, opened 2019-09-12
New dwm config.

GasparVardanyan:


noptrix:
@GasparVardanyan thanks for this mate. but as i stated in discord, we've removed dwm(-support) from BA.

Issue #2474, opened 2019-09-24
VMCloak - missing dependencies


Brueggus:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description
After installing vmcloak from the blackarch repository, I get the following errors:

```
pkg_resources.DistributionNotFound: The 'alembic<1.1,>=1.0.7' distribution was not found and is required by VMCloak
pkg_resources.DistributionNotFound: The 'sqlalchemy==1.3.3' distribution was not found and is required by VMCloak
pkg_resources.DistributionNotFound: The 'pyyaml==3.12' distribution was not found and is required by VMCloak
```

### Steps to reproduce
- pacman -S vmcloak
- vmcloak init --verbose --win7x64 win7x64base --cpus 2 --ramsize 4096

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

I get the following errors:

```
pkg_resources.DistributionNotFound: The 'alembic<1.1,>=1.0.7' distribution was not found and is required by VMCloak
pkg_resources.DistributionNotFound: The 'sqlalchemy==1.3.3' distribution was not found and is required by VMCloak
pkg_resources.DistributionNotFound: The 'pyyaml==3.12' distribution was not found and is required by VMCloak
```

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

I do not get the errors mentioned above.

### Info for developers

**Device:** Manufacturer Model XVI (Only if you consideer needed, for example, network cards)
**GNU/Linux distribution:** (If you are using one distro that is not BlackArch, example: Manjaro, Anarchy, etc)
**Tool version:** vmcloak (2:0.4.8-1)

noraj:
- VMCloak PKGBUILD: https://github.com/BlackArch/blackarch/blob/master/packages/vmcloak/PKGBUILD
- VMcloak upstream requirements.txt: https://github.com/hatching/vmcloak/blob/master/requirements.txt

I should try it with blackarch-devtool and see.

Brueggus:
Yes... the requirements.txt seems to be outdated. :|

https://github.com/hatching/vmcloak/blob/master/setup.py might help. They seem to be enforcing specific versions though.

noraj:
[...]

PR #2481, opened 2019-10-02
[WIP] vmcloak: fix PKGBUILD

noraj:
see #2474
WIP because install still failing

### DIFF
--- a/packages/vmcloak/PKGBUILD
+++ b/packages/vmcloak/PKGBUILD
@@ -10,13 +10,16 @@ url='https://github.com/jbremer/vmcloak'
 arch=('any')
 license=('GPL3')
 depends=('python2' 'cdrkit' 'virtualbox' 'dmidecode' 'lshw' 'python2-lockfile'
-         'python2-sphinx')
+         'python2-sphinx' 'python2-click' 'python2-jinja' 'python2-pyaml'
+         'python2-sqlalchemy' 'python2-alembic' 'python2-pefile')
 makedepends=('python2-setuptools' 'unzip')
 source=("https://github.com/jbremer/vmcloak/archive/$pkgver.tar.gz")
 sha512sums=('bc66e64ceed4b35eb1fbfe82cef8bd4a4024571ed66ecaadb247494b91fcfd4cde1f9ff0183148dc654da6a6c1b5061b7e17ab36aeffbf67080ea45f6afd5d8a')
 
 build() {
   cd "$pkgname-$pkgver"
+
+  sed -i -E 's/[=>]=\S[0-9.,< ]+//' setup.py
 
   python2 setup.py build
 }
--- a/packages/vmcloak/PKGBUILD
+++ b/packages/vmcloak/PKGBUILD
@@ -3,14 +3,13 @@
 
 pkgname=vmcloak
 pkgver=0.4.8
-pkgrel=1
-epoch=2
+pkgrel=2
 pkgdesc='Automated Virtual Machine Generation and Cloaking for Cuckoo Sandbox.'
 groups=('blackarch-malware')
 url='https://github.com/jbremer/vmcloak'
 arch=('any')
 license=('GPL3')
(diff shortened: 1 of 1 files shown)


Edu4rdSHL:
@noraj you need to patch `setup.py` and remove the static versions required here: https://github.com/hatching/vmcloak/blob/master/setup.py#L42-L58. They should look:

```
install_requires=[
        "click",
        "jinja2",
        "pefile2",
        "pyyam",
        "sqlalchemy",
        "alembic",
    ],
...
```

noraj:
It shouldn't mater as we are installing dependencies in a virtual environment `python setup.py install --root="$pkgdir" --prefix=/usr` so they can bring exact version if they want to. Removing the version requirement will work but I wonder if some are not respected they could break the app.

Edu4rdSHL:
It matter in all aspects. In the PKGBUILD we put the dependency name and it fetch the latest version in our repos.

When you install using the setup.py file with explicit versions numbers it make the version of the dependency a condition, that's why putting specific versions there make sense. If you look in out python packages, there are a lot of patches just to remove that thing.

noraj:
I don't understand what you mean. All dependencies listed in setup.py are not in `PKGBUILD`. `python setup.py install --root="$pkgdir" --prefix=/usr` will fetch the exact version required for the package and install them in a virtual folder out of the system python tree. So this don't fetch packages from our repo but from pypi nad we use a virtual folder exactly for not caring about dependencies requirements not being updated. If teh app use the dependencies install in the virtual environment which were exactly the ones required in setup.py I don't understand how it can fails. Please explain me on discord if I didn't understand something.

Edu4rdSHL:
@noraj please can you just patch the setup file and try again? You can check it yourself.

TL;DR:

When you build a Python package using specific versions it try to use the specific versions when installed, if you don't use specific versions in setup.py it try to work with any version. Python packages using setup.py doesn't download anything using pip, they use system packages (that's why they are included in depends=().

noraj:
I pushed the fix but the problem is that the BA package [python2-pefile] seems to match pypi [pefile](https://pypi.org/project/pefile/) but setup.py requires pypi [pefile2](https://pypi.org/project/pefile2/) instead are form the same repo only old releases like 1.2.11 (before 2013) after that date the releases changed to year.month.day format.
Since it isn't a different package but only a old version of an existing one I'll open an upstream bug.

noraj:
@Edu4rdSHL as all dependencies are very outdated anyway and upstream won't upgrade anytime soon we may need to switch to a [pip](https://wiki.archlinux.org/index.php/Python_package_guidelines#pip) install as we are already doing with gem for all ruby packages.

Edu4rdSHL:
We should remove the package instead or wait for upstream changes, using pip to install things as root is one of the worse things I've seen and will only cause most problems that solving. 

What happens if you change the dependency from pefile2 to pefile instead? Or well, we can package a python2-pefile2 package as well.

noraj:
I said **all the dependencies are very outdated**. They all have one or several MAJOR version late with several breaking changes not only for pefile but for all of them. So there is no way it will work with latest version of all of them.

For  ruby we install all the dependencies in a virtual vendor folder under `/usr/share/$pkgname/`, I think it is possible to do the same with python. I never thought of using pip to install packages globally with the system ones.

noraj:
Something like

```
pip install -r requirements.txt -t vendor
```

As it's done for composer or npm by default or for ruby with `bundle install --path vendor/bundle`

https://stackoverflow.com/questions/52695451/install-python-package-at-current-directory

mrsmith0x00:
@noraj is it done for merge?

noraj:
@mrsmith0x00 no I gave up

noptrix:
hey guys. i fixed this package and did a version bump (switched to git master branch, vcs version): https://github.com/BlackArch/blackarch/commit/d4447a52431d903ba3a40edb7a57256b69f0b273

PR #2542, opened 2019-11-21
multiple tools addition and update

noraj:
ruby-docopt: add
haiti: update
haiti-git: add
rabid: update
rabid-git: add
ruby-ctf-party: add
ruby-ctf-party-git: add

I made a **proper** packaging for ruby gems following **archlinux standards**.
The `--ignore-dependencies` option of gem is making sure that no dependencies are installed like when you do a `python setup.py install` for a python package.

I made both a release version and a bleeding edge git version.


noraj:
yep and groups are missing

noraj:
I'm done for groups and final newlines.
I added https://editorconfig.org/ to help enforcing final newlines in PKGBUILD.

Edu4rdSHL:
Please fix the Travis. You don't need to use `local` there, just `var_name=something`. And please check 9th line.

Edu4rdSHL:
Thank you a lot @noraj.

noptrix:
@noraj first thanks for your work. but guys: we try to not offer ruby-* packages whenever possible, because it's a pain with ruby gems. that's why we have *.install files where we use 'bundle'. second of all, we don't flag and provide -git packages. instead we'll soon have a 'blackarch-stable.db' where we move the release versions to that db and leaving 'blackarch.db' for -vcs versions. so ppl can decide whether to use vcs or stable versions. leave this decision to the user then. i will follow up on this soon after the ISO releases and let you know

noraj:
@noptrix `ruby-docopt` is a dependency of other tools, and `ruby-ctf-party` is my CTF lib. I can handle them. And with my new packaging method (actually the arch standard), packaging gems is easy.
Using `gem install --ignore-dependencies` is actually the same as `python setup.py install`.

And I know we don't provides `-git` but as I knew you were about to made create a stable repo I wanted to package both.

noptrix:
i cannot find a single blackarch package depending on 'ruby-docopt'. r u sure? we are using *.install files combined with bundler. yeah, fair enough go with 'ruby-ctf-party'. the opposite was a pain when we had to deal with all ruby-* packages. since we use the vendor dir / bundle less packages were broken. most of the ruby tools do work now... let's discuss this in IRC. if there is a special gem you want to release, like for example ctf-party then it's ok. but we are not going to add ruby-* dependencies for blackarch tools. see ya in irc

PR #2582, opened 2019-12-23
update neo4j-community package

cerberusec:
...


cerberusec:
the pkgcheck script seems to be throwing false positives. Not sure what it's getting tripped up on.

PR #2584, opened 2019-12-27
update to 3.5.14

cerberusec:
shortly after my last commit, I realized that update 3.5.14 was available. Note - FP's with pkgcheck again


Edu4rdSHL:
Thanks.

PR #2587, opened 2019-12-30
update powersploit pkgbuild

cerberusec:
went down a multi-day rabbit hole because powersploit was not on the 'dev' branch :P

Also updated the source url


noptrix:
@cerberusec thank you!

PR #2599, opened 2020-01-08
weevely: upgrade to python3

noraj:


Edu4rdSHL:
The package versión is still the same?

noptrix:
wow nice thanks for this @noraj. yeah @Edu4rdSHL i will bump

noptrix:
@Edu4rdSHL btw no, the versio is correct but we missed to switch to py3 and the current version in BA repo still uses py2 simply. so it's correct here just to pkgrel. thanks guys

Edu4rdSHL:
No problem, I was just thinking why the package was bumped from python2 to python3 without a pkgver bump.

> but we missed to switch to py3 and the current version in BA repo still uses py2 simply.

I suspected that was happening. Thanks for clarifying.

PR #2618, opened 2020-01-26
node-ar-drone: change dependencies

noraj:
npm seems more like build dependency than a runtime dependency


Edu4rdSHL:
Yeah, it's just building. Can you add the package to `lists/to-release`?

noraj:
I did the move for 2 other packages.

PR #2684, opened 2020-03-09
UhOh365: Initial commit

sachinkamath:
This PR adds PKGBUILD for https://github.com/Raikia/UhOh365

### DIFF
--- a/packages/uhoh365/PKGBUILD
+++ b/packages/uhoh365/PKGBUILD
@@ -1,0 +1,34 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=uhoh365
+pkgver=24.bc22611
+pkgrel=1
+pkgdesc='Script to enumerate Office 365 users without performing login attempts'
+groups=('blackarch' 'blackarch-recon')
+arch=('any')
+url='https://github.com/Raikia/UhOh365'
+license=('GPL-3.0')
+depends=('python' 'python-requests')
+makedepends=('git')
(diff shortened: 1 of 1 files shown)


Edu4rdSHL:
Hello, thanks. Have you tested it PKGBUILD? Because looking so far the pkgname is lowercase but when you are cloning the repo it creates a folder with lower and uppercase letters, so the cd $pkgname will not work.

packages/uhoh365/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+pkgver=24.bc22611
+pkgrel=1
+pkgdesc='Script to enumerate Office 365 users without performing login attempts'
+groups=('blackarch' 'blackarch-recon')
+arch=('any')
+url='https://github.com/Raikia/UhOh365'
+license=('GPL-3.0')
+depends=('python' 'python-requests')
+makedepends=('git')
+source=('git+https://github.com/Raikia/UhOh365.git')
```
Edu4rdSHL: Use `"git+https://github.com/Raikia/$pkgname.git"`.

sachinkamath:
> Hello, thanks. Have you tested it PKGBUILD? Because looking so far the pkgname is lowercase but when you are cloning the repo it creates a folder with lower and uppercase letters, so the cd $pkgname will not work.

I had it handled in prepare() but looks like `_pkgname` is the way to go :)

Will keep it innmind.

Edu4rdSHL:
Hey man, yes, in that case you should use `_pkgname`, very practical without need to do unneeded steps. Thank you for the PR man, we really appreciate it and hope you continue doing it.

PR #2685, opened 2020-03-10
drozer: Fixed typo in PKGBUILD.

ardevd:
The PKGBUILD fails to install due to a upper case typo. Fixed that and tested that it builds correctly.


Edu4rdSHL:
Hey man, nice catch! Thanks.

PR #2708, opened 2020-03-21
wapiti: remove useless dep

noraj:
'python-setuptools' needed on on make dep


Edu4rdSHL:
Hello, why do you consider it as useless dependency? There are two `setup.py` sections in the PKGBUILD (L22 and L28).

https://github.com/BlackArch/blackarch/blob/75f7a9903dcce2c64197eee25327543aa67c9697/packages/wapiti/PKGBUILD#L19-L32

Sorry but it will not be merged.

noraj:
@Edu4rdSHL I misswrote my PR comment. I wanted to say that 'python-setuptools' is only required in `makedepends` but not in `depends`. Actually it is in both.

https://github.com/BlackArch/blackarch/blob/75f7a9903dcce2c64197eee25327543aa67c9697/packages/wapiti/PKGBUILD#L11-L14

Edu4rdSHL:
Thank you @noraj!

Issue #2647, opened 2020-02-10
[Meta] add AWS tools


noraj:
- [ ] https://github.com/RhinoSecurityLabs/pacu

Many other tools: https://github.com/toniblyx/my-arsenal-of-aws-security-tools

noptrix:
added

PR #2726, opened 2020-04-06
Pacu: Initial commit

sachinkamath:
This PR adds [pacu](https://github.com/RhinoSecurityLabs/pacu) 

Should check off an item from #2647


noptrix:
@sachinkamath this is awesome. thanks a lot!

PR #2732, opened 2020-04-09
Add python-distro as a dependency for mobsf

defektive:
Currently when installing mobsf on a clean install of blackarch, I receive a handful of times.
```
ModuleNotFoundError: No module named 'distro'
```

This should prevent that.


noptrix:
@defektive thank you!

PR #2749, opened 2020-04-30
Add pspy to release

aancw:
```
pspy - version:  - Commit SHA:


     ██▓███    ██████  ██▓███ ▓██   ██▓
    ▓██░  ██▒▒██    ▒ ▓██░  ██▒▒██  ██▒
    ▓██░ ██▓▒░ ▓██▄   ▓██░ ██▓▒ ▒██ ██░
    ▒██▄█▓▒ ▒  ▒   ██▒▒██▄█▓▒ ▒ ░ ▐██▓░
    ▒██▒ ░  ░▒██████▒▒▒██▒ ░  ░ ░ ██▒▓░
    ▒▓▒░ ░  ░▒ ▒▓▒ ▒ ░▒▓▒░ ░  ░  ██▒▒▒
    ░▒ ░     ░ ░▒  ░ ░░▒ ░     ▓██ ░▒░
    ░░       ░  ░  ░  ░░       ▒ ▒ ░░
                   ░           ░ ░
                               ░ ░

Usage:
  pspy [flags]

Flags:
  -c, --color                        color the printed events (default true)
      --debug                        print detailed error messages
  -d, --dirs stringArray             watch these dirs
  -f, --fsevents                     print file system events to stdout
  -h, --help                         help for pspy
  -i, --interval int                 scan every 'interval' milliseconds for new processes (default 100)
  -p, --procevents                   print new processes to stdout (default true)
  -r, --recursive_dirs stringArray   watch these dirs recursively (default [/usr,/tmp,/etc,/home,/var,/opt])
t
```


noraj:
@aancw As pspy is more meant to be deployed on a compromised machine than ran on our own machine, I think embedding also 32 bits and bot the static and small binaries would be nice. So one can copy it to upload on the target machine.

aancw:
I'm thinking about that too, but as discussed before that I need to compile it too. It would be considered for change.

packages/pspy/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+  GOPATH="$srcdir" go get -d -t .
+}
+
+package() {
+  cd $pkgname
+
+  GOPATH="$srcdir" go build -v .
+
+  install -Dm 755 pspy "$pkgdir/usr/bin/$pkgname"
+  install -dm 755 "$pkgdir/usr/share/$pkgname/"{images,docker}
```
noraj, line 35: Why are you packaging this https://github.com/DominicBreuker/pspy/tree/master/images ? For the doc ? so you should move the images folder in /usr/share/doc/$pkgname/ too.

noraj: And i don't remember us packaging the dockerfiles too.

packages/pspy/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
+
+build() {
+  cd $pkgname
+
```
noraj, line 25: I don't know about go, just wondering if it is normal that https://github.com/BlackArch/blackarch-pkgbuilds/blob/08bfbf4a186ff2d4e5784ae06d83cc3f611d5b73/PKGBUILD-go#L29-L30 is not there.

aancw: Reference to this: 
https://github.com/BlackArch/blackarch/blob/master/packages/gau/PKGBUILD

packages/pspy/PKGBUILD
aancw: My bad. Thanks.

packages/pspy/PKGBUILD
aancw: Which one is correct?

cc @noptrix

noptrix: @aancw both are ok.

noptrix:
@aancw thanks a lot for this.

aancw:
@noptrix I've pushed new update for this pkgbuild. So it following the Go PKGBUILD.

aancw:
Please review my commit :) https://github.com/BlackArch/blackarch/commit/25e7f03f255f4f4ea399acaa81cbb4ed66bd0ff7

packages/pspy/PKGBUILD
aancw: Resolved in new PKGBUILD and noted for future.

PR #2760, opened 2020-05-06
[WIP] amoco: fix package

noraj:
- [x] python-pytest: runtime -> make dep
- [x] add missing dep for new version
- [x] python-ccrawl package not existing yet
- [ ] pyside2 dependency issue that made the package broken
    - `pkg_resources.DistributionNotFound: The 'PySide2' distribution was not found and is required by the application`
    - `pyside2 python-pyside python-pyside-common` are installed -> need to investigate
- [x] fix pkgver to use last tag
- [x] remove install for GPL2 license ([why](https://wiki.archlinux.org/index.php/PKGBUILD#license))

[upstream setup.py](https://github.com/bdcht/amoco/blob/release/setup.py)

### DIFF
--- a/packages/amoco/PKGBUILD
+++ b/packages/amoco/PKGBUILD
@@ -2,32 +2,25 @@
 # See COPYING for license details.
 
 pkgname=amoco
-pkgver=v2.4.1.r264.g5352aea
-pkgrel=2
-epoch=1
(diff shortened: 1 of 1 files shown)


noptrix:
@noraj i've tested it. this seems to be fixed. feel free to re-open if wrong.

PR #2762, opened 2020-05-11
Update snort.service

jaykijay:
Update Snort.service file to snort@.service so that Interface can be given via \@ as it's requested by %I in the service file.
Example: systemctl start snort@eth0.service


aancw:
It's already an alias there: https://github.com/BlackArch/blackarch/blob/master/packages/snort/snort.service

Edu4rdSHL:
Hello, @jaykijay thanks. Please can you bump the pkgrel?

@aancw that alias doesn't make the service to accept parameters in the way `service@parameter.service`. Take a look to systemd [instanced service units](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Default%20Dependencies).

Edu4rdSHL:
Thanks. Looks like we need to fix a few more things in [build](https://travis-ci.org/github/BlackArch/blackarch/builds/685706154#L1264). If you want to continue helping us we will appreciate this, if not I will fix this in a few hours!

PR #2770, opened 2020-05-22
add ActiveDirectoryEnum

noraj:


noraj:
@noptrix What's the better?

`/usr/bin/activedirectoryenum` (so the [CI stop complaining](https://travis-ci.org/github/BlackArch/blackarch/builds/690044674)) or `/usr/bin/activeDirectoryEnum` which is more readable and is named as source script.

aancw:
I've not seen camel case style for `/usr/bin/` file. `activedirectoryenum` is standart file name, or if you want more readable separate it with `-` `active-directory-enum`

noptrix:
@noraj thanks for this mate! btw, exactly what @aancw said, we either use all lowercase and/or `foo-bar-baz` style for /usr/bin/ files.

PR #2777, opened 2020-06-08
add vbsmin

noraj:
add [vbsmin](https://noraj.github.io/vbsmin/), vbs minifier


packages/vbsmin/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+  cd $pkgname
+
+  git describe --long | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+# waiting for ruby-paint 2.2.0 to be released on AL repo
+prepare() {
+  cd $pkgname
+
+  sed -i "s/'paint', '~> 2.2'/'paint', '>= 2.1'/" "$pkgname.gemspec"
```
Edu4rdSHL, line 28: Use `sed -i 's/~> 2.2/>= 2.1/g'`

noraj: Shorter but less future proof and anyway will be remove when https://www.archlinux.org/packages/community/any/ruby-paint/ will be updated

Edu4rdSHL: Don't open PRs then if you're going to omit comments, push directly to master.

PR #2785, opened 2020-06-14
snowman: pkgver bump

noraj:
tested with ba-dev :heavy_check_mark:


noraj:
@noptrix seems the CI is borken

PR #2795, opened 2020-06-23
Update pwncat from 0.0.23-alpha to 0.1.0

cytopia:
# Update pwncat from 0.0.23-alpha to 0.1.0

* https://github.com/cytopia/pwncat
* https://pwncat.org


cytopia:
Additionally I had a look at how ArchLinux is doing the package and they're using pip to install it instead:

```ruby
build() {
  cd "${pkgname}-${_pkgver}"
  python setup.py build
  gzip --force man/pwncat.1
}

package() {
  cd "${pkgname}-${_pkgver}"
  python setup.py install --root="$pkgdir" --optimize=1 --skip-build
  install -Dm644 LICENSE.txt "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
  install -Dm644 *.md -t "${pkgdir}/usr/share/doc/${pkgname}"
  install -Dm644 man/pwncat.1.gz -t "${pkgdir}/usr/share/man/man1"
}
```

https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=pwncat

Do you want me to change the install routine from simple copying to `pip` as well?

packages/pwncat/PKGBUILD
```diff
@@ -11,11 +11,11 @@ url='http://pwncat.org/'
 license=('MIT')
 arch=('any')
 depends=('python')
```
cytopia, line 13: Should this be changed to `python3`?

aancw:
> Do you want me to change the install routine from simple copying to `pip` as well?

As far as I know, BA don't use pip for installing package, use that python `setup.py` instead.

cytopia:
> As far as I know, BA don't use pip for installing package, use that python `setup.py` instead.

Yes you're right, that's what I actually meant (instead of `install ...`). I will update my post above to reflect this.

aancw:
> Yes you're right, that's what I actually meant (instead of `install ...`). I will update my post above to reflect this.

I don't see any problem with your new PKGBUILD in BA or AUR. Both new version or old is okay, but the new version use source instead of binary. BA prefer using source when installing package.

Ref: https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python

aancw:
Btw, Thanks for creating an awesome tool for offensive purpose. I'm fallin in love with your tool feature. Long live pwncat （๑✧∀✧๑）

packages/pwncat/PKGBUILD
Edu4rdSHL: No.

Edu4rdSHL:
Hello, thanks! Please add the tool name to `lists/to-release`.

Edu4rdSHL:
Thank you, @cytopia!

PR #2804, opened 2020-07-15
Fix change directory error

rmasp98:
Scoutesuite by default writes files in your current directory. As this package cd's to the package directory, it returns a permission error as well as not writing to the expected directory. I have removed the cd and referenced the correct directory in both the source and the python execution

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -2,3 +2,4 @@ python-typing
 python-hstspreload
 python-graphviz
 python-azure-cli-core
+scoutsuite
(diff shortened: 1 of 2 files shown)


noraj:
@rmasp98 Yes classic issue. thanks for your PM. LGTM but did you tested it?

rmasp98:
I can confirm I used scoutsuite today with the changes and seemed to work fine but have not done any particularly thorough testing.

@Edu4rdSHL added requested changes but double check I have done what you asked as not done this stuff before

Edu4rdSHL:
@rmasp98 thank you.

Issue #2812, opened 2020-07-23
Java bytecode editors & decompilers


noraj:
- [x] Cafebabe #2815
  - https://grax.info/
  - https://github.com/GraxCode/Cafebabe
- [ ] ~~JByteMod #2816~~ tool is abandoned and not building
  - https://github.com/GraxCode/JByteMod-Beta
- [x] JBE
  - https://set.ee/jbe/
- [ ] JBytedit
  - https://github.com/Eyremba/JBytedit
- [x] reJ
  - http://rejava.sourceforge.net/

D3vil0p3r:
#4226

D3vil0p3r:
[reJ #4228](https://github.com/BlackArch/blackarch/pull/4228)

D3vil0p3r:
This issue can be closed. [JByteMod is an abandoned project in favor of Cafebabe](https://github.com/GraxCode/JByteMod-Beta).

PR #2816, opened 2020-07-24
[WIP] add jbytemod

noraj:
- [ ] Need to wait for this upstream issue: https://github.com/GraxCode/JByteMod-Beta/issues/86

related to #2812

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+jbytemod
--- a/packages/jbytemod/PKGBUILD
+++ b/packages/jbytemod/PKGBUILD
@@ -1,0 +1,45 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=jbytemod
+pkgver=1.8.2.285.57e6f8e
+pkgrel=1
+groups=('blackarch' 'blackarch-decompiler' 'blackarch-reversing')
+pkgdesc='Java bytecode editor & decompiler.'
+arch=('any')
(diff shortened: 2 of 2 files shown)


noptrix:
@noraj any news on this?

ikstream:
Just looked into the tool and there is an open PR that addresses the issue. Seems like an abandoned tool. We could apply the PR as a patch to fix the tool in our repo

noraj:
> @noraj any news on this?

Nop, abandoned since 2019.

> Just looked into the tool and there is an open PR that addresses the issue. Seems like an abandoned tool. We could apply the PR as a patch to fix the tool in our repo

Yes if we don't have to maintain it after that.

ikstream:
> > Just looked into the tool and there is an open PR that addresses the issue. Seems like an abandoned tool. We could apply the PR as a patch to fix the tool in our repo
> 
> Yes if we don't have to maintain it after that.

As it is abandoned since 2019 I don't expect any maintenance overhead with a possible patch

noraj:
> > > Just looked into the tool and there is an open PR that addresses the issue. Seems like an abandoned tool. We could apply the PR as a patch to fix the tool in our repo
> > 
> > 
> > Yes if we don't have to maintain it after that.
> 
> As it is abandoned since 2019 I don't expect any maintenance overhead with a possible patch

If the fix is about updating dependencies (https://github.com/GraxCode/JByteMod-Beta/pull/88), it means will have to do it from time to time or am I wrong?

noraj:
tool abandonned

PR #2873, opened 2020-10-17
cmsscan: add tool

### LINKED ISSUE #2244: [Tool request] add CMSScan
- [x] I have searched open and closed issues for duplicates

https://github.com/ajinabraham/CMSScan

noraj:
fix #2244

### DIFF
--- a/packages/cmsscan/PKGBUILD
+++ b/packages/cmsscan/PKGBUILD
@@ -1,0 +1,54 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+_pkgname=CMSScan
+pkgname=cmsscan
+pkgver=41.45a0360
+pkgrel=1
+pkgdesc='CMS scanner to identify and find vulnerabilities for Wordpress, Drupal, Joomla, vBulletin'
+arch=('any')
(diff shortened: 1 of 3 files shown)


packages/cmsscan/PKGBUILD
```diff
@@ -0,0 +1,54 @@
...
+
+package() {
+  cd $_pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+  install -dm 755 "$pkgdir/usr/share/$pkgname"
+
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
+
+  rm -rf *.md LICENSE *.sh *.txt plugins/joomscan plugins/vbscan LICENSES
```
Edu4rdSHL, line 42: Hey @noraj, are you sure that it's necessary to delete the `plugins/*` folders? I guess we should keep them just in case the tool wants to save/use something from there.

noraj: @Edu4rdSHL plugins/joomscan plugins/vbscan and git submodules but we don't need them because we are using the system packaged depencies rather than the git submodules see https://github.com/ajinabraham/CMSScan/pull/26/files

noraj: That's for other distro where those tools are not packaged and that will use [setup.sh](https://github.com/ajinabraham/CMSScan/blob/master/setup.sh)

PR #2874, opened 2020-10-17
Add vipermonkey

pwnfoo:
This PR adds [ViperMonkey](https://github.com/decalage2/ViperMonkey/) and fixes #2572

PR #2875, opened 2020-10-17
vipermonkey: add missing install and patch files

pwnfoo:
Forgot this in #2874  sorry :(


Edu4rdSHL:
I doubt the tool will work this way as we have `python2-pyparsing 2.4.7-1` in the repos, I think that when you try to launch the tool it will fails due to the specify version of `python2-pyparsing` in `setup.py`.

pwnfoo:
@Edu4rdSHL That's why I added virtualenv in the package. It works for me.

image

PR #2876, opened 2020-10-17
organised the packages and added a new version of hulk

Chr0m0s0m3s:
Organised the packages with their respective category folders , and added the improved version of hulk python3 and golang with the tools.
The new one has the following capacities:
1> taking down websites hosted on cloudflare if not configured properly
2> bringing down sites with apache servers

Note :- if the site is poorly configured without CDN or loadbalancing it would take the whole backend server down else it will just take down the closest server to the user, which means that for other people trying to connect to the website might see it up, if they arent connected to the same server of the CDN as the user.


Edu4rdSHL:
Hello @Chr0m0s0m3s, thank you a lot for the idea, it's really good to keep packages separated according to their category, however, the current structure is used by our build system and we need to adapt it to the new one in case we merge this.

I'm converting this PR to draft first and I'll talk with the other team members to decide what to do.

Regards,
Eduard

Chr0m0s0m3s:
an example of the new structured folder can be found here : https://github.com/Chr0m0s0m3s/blackarch/tree/blackarch/packages
The new improved hulk tool can be found here : https://github.com/Chr0m0s0m3s/hulk

Edu4rdSHL:
As the owner of this PR has deleted their Github account, I'm closing this.

PR #2902, opened 2020-11-07
favfreak

### LINKED ISSUE #2888: [Tool Request] FavFreak
- **Name**: FavFreak
- **Source URL**: https://github.com/devanshbatham/FavFreak

dualfade:
close #2888

### DIFF
--- a/packages/favfreak/PKGBUILD
+++ b/packages/favfreak/PKGBUILD
@@ -1,0 +1,44 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=favfreak
+_pkgname=FavFreak
+pkgver=19.3a0bf2a
+pkgrel=1
+pkgdesc='Weaponizing favicon.ico for BugBounties , OSINT and what not.'
+arch=('any')
+groups=('blackarch' 'blackarch-recon' 'blackarch-fingerprint')
+url='https://github.com/devanshbatham/FavFreak'
+license=('custom:unknown')
+depends=('python' 'python-virtualenv')
+makedepends=('git' 'python-pip')
+source=("git+https://github.com/devanshbatham/$_pkgname.git")
+sha512sums=('SKIP')
+install="$pkgname.install"
+
+pkgver() {
+  cd $_pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
+
+package() {
+  cd $_pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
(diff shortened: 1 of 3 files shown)


packages/favfreak/favfreak.install
```diff
@@ -0,0 +1,13 @@
+post_install() {
+  set -e
+  cd /usr/share/favfreak
+  python -m virtualenv venv
```
noraj, line 4: virtualenv is reserved for as the last resort packaging way for complex software with tons or very unmaintained dependencies.

favfreak has only one dependency: [mmh3](https://pypi.org/project/mmh3/) and this dependency is pure python no sub-dependencies, so very easy to package.

Package favfreak with [this template](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python) and mmh3 with [this template](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib).

noraj (changes requested):

packages/favfreak/favfreak.install
dualfade: Ok Np will go to PKGBUILD only. I was using virtualenv install instructions from github.

packages/favfreak/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+
+pkgname=favfreak
+_pkgname=FavFreak
+pkgver=19.3a0bf2a
+pkgrel=1
+pkgdesc='Weaponizing favicon.ico for BugBounties , OSINT and what not.'
+arch=('any')
+groups=('blackarch' 'blackarch-recon' 'blackarch-fingerprint')
+url='https://github.com/devanshbatham/FavFreak'
+license=('custom:unknown')
```
noraj, line 12: When no license contacting upstream is important https://github.com/devanshbatham/FavFreak/issues/6

packages/favfreak/PKGBUILD
```diff
@@ -0,0 +1,35 @@
...
+depends=('python' 'python-virtualenv' 'python-mmh3')
+makedepends=('git' 'python-pip')
```
noraj, line 14:
```diff
-python-pip
+python-setuptools
```

noraj (changes requested):

packages/favfreak/PKGBUILD
dualfade: no update from from @devanshbatham as of yet. Will update once we have this.

dualfade: updated. Applying

PR #2906, opened 2020-11-08
python-ccrawl

dualfade:
https://github.com/BlackArch/blackarch/pull/2760

please review.

### DIFF
--- a/packages/python-ccrawl/PKGBUILD
+++ b/packages/python-ccrawl/PKGBUILD
@@ -1,0 +1,37 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=python-ccrawl
+pkgname=('python-ccrawl')
+_pkgname=ccrawl
+pkgver=1.2
+pkgrel=1
+pkgdesc='Ccrawl uses clang to build a database related to various C/C++ data structures.'
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
+url='https://github.com/bdcht/ccrawl'
+license=('GPL3')
+makedepends=('git' 'python-pip')
+source=("git+https://github.com/bdcht/$_pkgname.git")
+sha512sums=('SKIP')
+
+prepare() {
+  cp -a "$_pkgname"{,-2}
+}
+
+build() {
+  cd "$srcdir/$_pkgname"
+
+  python setup.py build
+}
+
+package_python-ccrawl() {
+  depends=('python' 'python-virtualenv' 'python-click' 'python-traitlets' \
+    'python-tinydb' 'clang')
+
+  cd "$_pkgname"
+
+  python setup.py install --prefix=/usr --root="$pkgdir" --optimize=1 \
+    --skip-build
+}
+
--- a/packages/python-ccrawl/PKGBUILD
+++ b/packages/python-ccrawl/PKGBUILD
@@ -1,23 +1,17 @@
 # This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
 # See COPYING for license details.
 
-pkgbase=python-ccrawl
-pkgname=('python-ccrawl')
+pkgname=python-ccrawl
 _pkgname=ccrawl
 pkgver=1.2
 pkgrel=1
-pkgdesc='Ccrawl uses clang to build a database related to various C/C++ data structures.'
+pkgdesc='Uses clang to build a database related to various C/C++ data structures.'
 arch=('any')
-groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
 url='https://github.com/bdcht/ccrawl'
 license=('GPL3')
-makedepends=('git' 'python-pip')
+makedepends=('git' 'python-setuptools')
 source=("git+https://github.com/bdcht/$_pkgname.git")
 sha512sums=('SKIP')
-
-prepare() {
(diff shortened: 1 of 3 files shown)


packages/python-mmh3/PKGBUILD
```diff
@@ -0,0 +1,37 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=python-mmh3
```
noraj, line 4: Yet another git branching issue

noraj (changes requested):

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=python-ccrawl
+pkgname=('python-ccrawl')
```
noraj, line 5: don't use parenthesis when there is only one element in the array

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=python-ccrawl
```
noraj, line 4: pkgbase used only when you do 2 in 1 package like py2 + py3 lib, so if only py3 no pkgbase

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+pkgname=('python-ccrawl')
+_pkgname=ccrawl
+pkgver=1.2
+pkgrel=1
+pkgdesc='Ccrawl uses clang to build a database related to various C/C++ data structures.'
```
noraj, line 9: Don't put the pkgname in the desc.

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
```
noraj, line 11: not groups for dependencies

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+url='https://github.com/bdcht/ccrawl'
+license=('GPL3')
+makedepends=('git' 'python-pip')
+source=("git+https://github.com/bdcht/$_pkgname.git")
+sha512sums=('SKIP')
+
+prepare() {
+  cp -a "$_pkgname"{,-2}
```
noraj, line 19: remove `{,-2}` since you don't package the py2 version

noraj: Also remove double quote when there is no no dashes or slashes

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+}
+
+build() {
+  cd "$srcdir/$_pkgname"
+
+  python setup.py build
+}
+
+package_python-ccrawl() {
```
noraj, line 28: put only `package`, `package_python-ccrawl` is when you have 2 in 1 package

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+  depends=('python' 'python-virtualenv' 'python-click' 'python-traitlets' \
```
noraj, line 29: It seems that `python-virtualenv` is not required but python-pymongo or python-requests are missing https://github.com/bdcht/ccrawl/blob/release/requirements.txt

Also maybe python-python-rapidjson and python-pyparsing https://github.com/bdcht/ccrawl/blob/bbb434bce370c2e42044225a6a91ba7b15e8d58e/setup.py#L28-L35

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+    'python-tinydb' 'clang')
+
+  cd "$_pkgname"
```
noraj, line 32: no double quotes here

noraj (changes requested):

packages/python-ccrawl/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+pkgname=('python-ccrawl')
+_pkgname=ccrawl
+pkgver=1.2
+pkgrel=1
+pkgdesc='Ccrawl uses clang to build a database related to various C/C++ data structures.'
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
+url='https://github.com/bdcht/ccrawl'
+license=('GPL3')
+makedepends=('git' 'python-pip')
```
noraj, line 14:
```diff
-python-pip
+python-setuptools
```

noraj (changes requested):

packages/python-ccrawl/PKGBUILD
dualfade: python-python-rapidjson pukes. just fyi.

packages/python-ccrawl/PKGBUILD
dualfade: This should be removed with no py2 deps ? If so; will resolve. Thanks.

PR #2904, opened 2020-11-07
python-mmh3

dualfade:

PR #2909, opened 2020-11-09
Python mmh3

dualfade:
please review.


noraj:
new version of #2904
PS : there is still favfreak included in this PR, I'll try to fix that locally

dualfade:
Thank you. I will try to be much more mindful of this. My apologies.

PR #2913, opened 2020-11-11
add python-mmh3

noraj:
replace & close  #2909, #2904, #2888

dep required for #2888


noraj:
@dualfade I committed on your behalf see https://github.com/BlackArch/blackarch/pull/2913/commits

Issue #2829, opened 2020-08-09
StringSifter 


noraj:
- **Name**: StringSifter 
- **Source URL**: https://github.com/fireeye/stringsifter

dualfade:
Will PR

dualfade:
In progress --
Errors with py deps during build --

(1/1) installing stringsifter                                                                                                                                               [##########################################################################################################] 100%
created virtual environment CPython3.8.6.final.0-64 in 254ms
  creator CPython3Posix(dest=/usr/share/stringsifter/venv, clear=False, global=False)
  seeder FromAppData(download=False, pip=latest, setuptools=latest, wheel=latest, via=copy, app_data_dir=/root/.local/share/virtualenv/seed-app-data/v1.0.1)
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
Collecting lightgbm==2.1.2
  Downloading lightgbm-2.1.2-py2.py3-none-manylinux1_x86_64.whl (730 kB)
Collecting numpy==1.17.1
  Downloading numpy-1.17.1.zip (6.5 MB)
Collecting scikit-learn==0.21.3
  Downloading scikit-learn-0.21.3.tar.gz (12.2 MB)
    ERROR: Command errored out with exit status 1:
     command: /usr/share/stringsifter/venv/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-fal9r75b/scikit-learn/setup.py'"'"'; __file__='"'"'/tmp/pip-install-fal9r75b/scikit-learn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' --no-user-cfg egg_info --egg-base /tmp/pip-pip-egg-info-j2pftitu
         cwd: /tmp/pip-install-fal9r75b/scikit-learn/

dualfade:
On hold for now. Too many numpy dep errors. Multi version.

PR #2918, opened 2020-11-12
StringSifter #2829

dualfade:
please review. Thank you.

### DIFF
--- a/packages/stringsifter/PKGBUILD
+++ b/packages/stringsifter/PKGBUILD
@@ -1,0 +1,57 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=stringsifter
+pkgver=14.f4331a0
+pkgrel=1
+pkgdesc='Machine learning tool that automatically ranks strings based on their relevance for malware analysis.'
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
+url='ihttps://github.com/fireeye/stringsifter'
+license=('Apache')
+depends=('python' 'python-virtualenv' 'python-numpy' 'python-scikit-learn' \
+  'python-joblib' 'python-pytest')
+makedepends=('git' 'python-setuptools')
+source=("git+https://github.com/fireeye/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
+
+build() {
+  cd $pkgname
+
+  python setup.py build
+}
+
+package() {
+  cd $pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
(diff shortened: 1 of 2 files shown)


packages/stringsifter/PKGBUILD
```diff
@@ -0,0 +1,57 @@
...
+
+pkgname=stringsifter
+pkgver=14.f4331a0
+pkgrel=1
+pkgdesc='Machine learning tool that automatically ranks strings based on their relevance for malware analysis.'
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
+url='ihttps://github.com/fireeye/stringsifter'
+license=('Apache')
+depends=('python' 'python-virtualenv' 'python-numpy' 'python-scikit-learn' \
```
noraj, line 12: python-virtualenv is not used please remove it

packages/stringsifter/PKGBUILD
```diff
@@ -0,0 +1,57 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=stringsifter
+pkgver=14.f4331a0
+pkgrel=1
+pkgdesc='Machine learning tool that automatically ranks strings based on their relevance for malware analysis.'
+arch=('any')
+groups=('blackarch' 'blackarch-binary' 'blackarch-forensic')
+url='ihttps://github.com/fireeye/stringsifter'
```
noraj, line 10: there is a `i` before the URL

packages/stringsifter/PKGBUILD
```diff
@@ -0,0 +1,57 @@
...
+license=('Apache')
+depends=('python' 'python-virtualenv' 'python-numpy' 'python-scikit-learn' \
+  'python-joblib' 'python-pytest')
```
noraj, line 13: `python-pytest` is only for the tests that we remove and don't package so remove this dependency

packages/stringsifter/PKGBUILD
```diff
@@ -0,0 +1,57 @@
...
+
+build() {
+  cd $pkgname
+
+  python setup.py build
+}
+
+package() {
+  cd $pkgname
+
```
noraj, line 32: ## Setup.py

There is a setup.py file so why not just using `python setup.py install --root="$pkgdir" --prefix=/usr -O1 --skip-build` ?

## jointscript

Also if you use the jointscript method the `build()` method is not used.
You forgot to remove the `tests` folder.

noraj:
@noptrix you can leave me the reviews if you are busy :wink:

packages/stringsifter/PKGBUILD
dualfade: Good catch. Fixed.

packages/stringsifter/PKGBUILD
dualfade: removed.

packages/stringsifter/PKGBUILD
dualfade: removed

PR #2919, opened 2020-11-14
stringsifter #2829 #2918

dualfade:

PR #2920, opened 2020-11-15
Stringsifter fix

noraj:
close #2919, #2829
fix #2918


noraj:
We will be able to remove the prepare statement when upstream will have merge https://github.com/fireeye/stringsifter/pull/15

Issue #2926, opened 2020-11-21
`go-pie` to `go` migration


ejiek:
Type: Discussion

## Discussion

Upstream AL got `go-pie` [replaced](https://github.com/archlinux/svntogit-community/commit/df61d6d21724f62c6b028e4c72efb580aade6e1a#diff-3e341d2d9c67be01819b25b25d5e53ea3cdf3a38d28846cda85a195eb9b7203a) by regular `go` due to adopting `go-pie` config as the default one.

The reason I'm creating this - `grype` is fairly new ( packaged long after `go-pie` change) yet it still uses `go-pie`.
As does everything else but `snapd`.

Is there a good reason for it or just a matter of [updating the template](https://github.com/BlackArch/blackarch-pkgbuilds/pull/8)?

*ps. I'm not concerned about old packages but new ones do bother me.*

noptrix:
@ejiek matter of template. we stick to 'go-pie'.

PR #2927, opened 2020-11-21
grype: version bump.

ejiek:
And a bit more:

* #2926 is addressed
* versions are converted to github releases (from commit number)
  * `epoch` is introduced to allow updates
* go build flags & options


ejiek:
I'm confused by build happening in `package`.
If there are no arguments for it, I'd like to build in `build`

noptrix:
thanks a lot @ejiek

PR #2938, opened 2020-12-06
Duplicut

noraj:


noptrix:
@noraj let's use ```blackarch-misc``` at least as primary group. it's rather a helper for crackers. also, ```to-release``` was just updated so you can update your PR.

noraj:
> @noraj let's use `blackarch-misc` at least as primary group. it's rather a helper for crackers. also, `to-release` was just updated so you can update your PR.

blackarch-cracking would have been a better group name

PR #2939, opened 2020-12-06
blackarch-cryptography -> blackarch-crypto

noraj:
blackarch-cryptography & blackarch-crypto are redundant/duplicate
since there were only 4 in blackarch-cryptography and about one hundred
in blackarch-crypto I moved blackarch-cryptography -> blackarch-crypto


noraj:
```
$ pacman -Sg | grep blackarch | grep crypto
blackarch-crypto
blackarch-cryptography
```

noptrix:
@noraj thanks a lot. ```cryptography``` was wrong anyway and is not listed in ```lists/groups```.

Issue #2925, opened 2020-11-21
enum4linux-ng


pwnfoo:
- **Name**: enum4linux-ng
- **Source URL**: https://github.com/cddmp/enum4linux-ng
- **Description**:  A next-generation version of enum4linux with additional features like JSON/YAML export. Aimed for security professionals and CTF players.

This one seems to have nicer features. Do you think this one is worth adding?

noraj:
enum4linux is unmaintained and buggy as hell so enum4linux-ng is welcome

noraj:
- [ ] setup.py https://github.com/cddmp/enum4linux-ng/issues/11

PR #2952, opened 2020-12-12
add enum4linux-ng

noraj:
close https://github.com/cddmp/enum4linux-ng/issues/11
close https://github.com/BlackArch/blackarch/issues/2925

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+enum4linux-ng
--- a/packages/enum4linux-ng/PKGBUILD
+++ b/packages/enum4linux-ng/PKGBUILD
@@ -1,0 +1,33 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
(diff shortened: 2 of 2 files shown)


Edu4rdSHL (approved):

Edu4rdSHL:
Hey @noraj, thank you! I only wonder if `enum4linux-ng` doesn't have any kind of conflicting files with `enum4linux` so that we need to add `conflicts=()` to the PKGBUILD.

cddmp:
@Edu4rdSHL 
There should are no conflicts. enum4linux is Perl while this one is Python. Both don't share files or interact in any way with each other.

Edu4rdSHL:
Hello, @cddmp, thank you for answering so quickly. Merging.

noraj:
Yes no conflict Python/perl + the way it is package even if both wold have been in python there would have been installed in a separate folder with a separate binary name so no worry.

cddmp:
Thanks!

PR #2970, opened 2021-01-04
version bump

noraj:
all vcs package using vcschecker


noptrix:
@noraj thanks for this, but we use this in another way and manual interaction is here and there required for some tools. i am closing this and you don't need to re-run it. we do it at least one time per week using upgrade-git-packages script

PR #2975, opened 2021-01-06
njsscan lib dep

dualfade:
please review --

### DIFF
--- a/packages/python-sarif-om/PKGBUILD
+++ b/packages/python-sarif-om/PKGBUILD
@@ -1,0 +1,28 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-sarif-om
+_pkgname=sarif_om
+pkgver=1.0.4
+pkgrel=1
+pkgdesc='Classes implementing the SARIF 2.1.0 object model.'
+arch=('any')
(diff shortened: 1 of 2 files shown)


packages/python-sarif-om/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+_pkgname=sarif_om
+pkgver=1.0.4
+pkgrel=1
+pkgdesc='Classes implementing the SARIF 2.1.0 object model.'
+arch=('any')
+url='https://pypi.org/project/sarif-om/#files'
+license=('MIT')
+depends=('python')
+makedepends=('python-setuptools')
+source=("https://files.pythonhosted.org/packages/ba/de/bbdd93fe456d4011500784657c5e4a31e3f4fcbb276255d4db1213aed78c/$_pkgname-$pkgver.tar.gz")
```
noraj, line 14: take https://github.com/microsoft/sarif-python-om/archive/1.0.4.tar.gz so the link doesn't change at each release

noraj (changes requested):

packages/python-sarif-om/PKGBUILD
noptrix: @dualfade no, keep pypi links since our script handles auto-upgrades :) sorry for the confusion

PR #2976, opened 2021-01-06
njsscan lib dep

dualfade:
please review --

### DIFF
--- a/packages/python-jschema-to-python/PKGBUILD
+++ b/packages/python-jschema-to-python/PKGBUILD
@@ -1,0 +1,28 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-jschema-to-python
+_pkgname=jschema_to_python
+pkgver=1.2.3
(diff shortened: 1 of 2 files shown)


packages/python-jschema-to-python/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+_pkgname=jschema_to_python
+pkgver=1.2.3
+pkgrel=1
+pkgdesc='.Generate Python classes from a JSON schema'
+arch=('any')
+url='https://pypi.org/project/jschema-to-python/#files'
+license=('MIT')
+depends=('python')
+makedepends=('python-setuptools')
+source=("https://files.pythonhosted.org/packages/1d/7f/5ae3d97ddd86ec33323231d68453afd504041efcfd4f4dde993196606849/$_pkgname-$pkgver.tar.gz")
```
noraj, line 14: maybe take https://github.com/microsoft/jschema-to-python/archive/1.2.3.tar.gz so the link does change at each release

noraj (changes requested):

Issue #2967, opened 2020-12-23
njsscan


vdun:
- **Name**: njsscan
- **Source URL**: https://github.com/ajinabraham/njsscan

Edu4rdSHL:
Added.

PR #2977, opened 2021-01-06
njsscan #2967

dualfade:
please review --


packages/njsscan/PKGBUILD
```diff
@@ -0,0 +1,30 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-njsscan
+_pkgname=njsscan
+pkgver=0.2.0
+pkgrel=1
+pkgdesc='A static application testing (SAST) tool that can find insecure code patterns in your node.js applications.'
+arch=('any')
+url='https://pypi.org/project/jschema-to-python/#files'
```
noraj, line 10: Wrong URL ? Is it https://pypi.org/project/njsscan/#files ?

packages/njsscan/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+license=('GPL2')
+depends=('python' 'python-libsast' 'python-tabulate' 'python-sarif-om' 'python-jschema-to-python'
+         'python-pbr' 'python-jsonpickle' 'python-libsast' 'python-semgrep' 'python-colorama' 'python-ruamel-yaml')
+makedepends=('python-setuptools')
+source=("https://files.pythonhosted.org/packages/7e/9e/072bf8c475ae75314e284892879b2dadbc29e305fdbffd3efadc8d3e1c3a/$_pkgname-$pkgver.tar.gz")
```
noraj, line 15: maybe prefer teh tar.gz from https://github.com/ajinabraham/njsscan/archive/0.2.0.tar.gz so the URL won't change at each release.

noraj (changes requested):

packages/njsscan/PKGBUILD
dualfade: Yes my bad. Adjusted. Have not checked in yet.

Issue #2901, opened 2020-11-02
nbtscan segmentation fault


saltasatelites:
- [x ] I have searched open and closed issues for duplicates
- [ x] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

Make any kind of query. You always get a segmentation fault. 

### Steps to reproduce

- install from blackarch or Archlinux/community. (same version 1.5.1-8)
- make a query. For example:  
 >nbtscan 192.168.1.25-137

- that reproduce the bug
- Segmentation fault
**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
  Scan a range of IP addresses

### Screenshots
none

### Info for developers

**GNU/Linux distribution:** Archlinux with Blackarch repository above. You can install from the community repo too. It's the same package.
**Tool version:** 1.5.1-8

### Link to debug log

No log.

ikstream:
ntbscan isn't maintained by Blackarch, as it is upstream. Please report to arch Linux, if it hasn't already be done

dualfade:
Will update build --

PR #2982, opened 2021-01-10
nbtscan #2901; pkgbuild src install

dualfade:
Please review; if we need to change pkgname to nbtscan-git please let me know. This package is compiling from src and seams to be running as exepcted.. Changelog included.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+nbtscan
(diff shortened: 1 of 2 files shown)


Edu4rdSHL:
Hello @dualfade , thank you for the PR. From my point of view its fine as it's, we have the package from official AL repos but our version is major due to the git versioning model and users only should view the update. Cc @noptrix

dualfade:
No worries. However you would like to proceed.

noptrix:
@dualfade @Edu4rdSHL thanks guys.

PR #3002, opened 2021-02-07
blackarch-keyring: Added pacman hook

net-wayfarer:
blackarch-keyring: Added pacman hook

Installs blackarch-key.hook into /etc/pacman.d/hooks so that the keys in
blackarch-keyring gets reinstated whenever archlinux-keyring is updated.

This should be a different commit compared to the mistakes that were
done in the past. Changes to remove unnecessary braces were implemented,
the changes are done on a separate branch, and here's hoping all should
work.


noptrix:
@net-wayfarer thanks a lot

PR #3016, opened 2021-02-13
Added The Working Repository Link

josh0xA:
The old repository doesn't exist anymore therefore PKGBUILD cannot actually install the tool. Fixed it by adding a forked version of Gloom-Framework.


Edu4rdSHL (changes requested):
Hello, @josh0xA. Thank you for the update, everything seems working fine but the new `pkgver` of the package is less than the one that existed before, please can you add `epoch` to the PKGBUILD? Also please add the package name to `lists/to-release`

Edu4rdSHL:
Thank you. @josh0xA!

Issue #3055, opened 2021-03-22
blackarch-menus


visious25:
- [ ] I have searched open and closed issues for duplicates
- [x ] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

blackarch-menus version 10.5244eea-1 it doesn't appears in xfce menu.

ikstream:
Hi, can you please check if `blackarch-menus` is installed?
If you installed it later, you would need to update tools you have installed prior, to make them show up

visious25:
Hi, I have installed blackarch-menus and nmap tool only. Before two versions i didn't have this problem.

Edu4rdSHL:
@visious25 if you have installed only `nmap`, how do you want to see all tools in the menu?

Actually the menu entries are created on the fly, so if you installed only one tool that menu entryis created. Now, I ask to you: the nmap entry exists?

PR #3056, opened 2021-03-22
blackarch-menus: add to blackarch group

ikstream:
We should add blackarch-menus to the blackarch group.
Currently if people run `pacman -S blackarch` it will not be installed
People get confused as the menu structure isn't showing up in desktop Environments that use `.desktop` files as it may be the case in #3055

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,1 +1,2 @@
 airgeddon
+blackarch-menus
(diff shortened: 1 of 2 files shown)


Edu4rdSHL (approved):
I agree with you, @ikstream.

PR #3064, opened 2021-04-02
c5scan: update mislabeled group.

wykwit:
Currently there is only one package in `blackarch-scan` group. I think the package was supposed to go into `blackarch-scanner` instead.


noraj:
Yes thanks, `scan` is not a valid group: `$ curl https://raw.githubusercontent.com/BlackArch/blackarch/master/lists/groups`

noraj (approved):

PR #3071, opened 2021-04-05
trufflehog cleanup

wykwit:
This package perfectly fits 'blackarch-code-audit' group.

Test suite packages ('python-codecov', 'python-pytest-cov') are not listed under truffleHog's `install_requires` and therefore can be excluded from `depends`.

I also think it's not necessary to install a common license which is already available under `/usr/share/licenses/common/` in Arch Linux.


noptrix:
@wykwit thanks a lot for this!

PR #3097, opened 2021-05-13
Add desktop file generator script

ikstream:

### DIFF
--- a/scripts/gen-desktop-files
+++ b/scripts/gen-desktop-files
@@ -56,7 +56,7 @@ get_name()
 get_desc()
 {
   pkg="$1"
-  comment="$(awk -F '=' '/pkgdesc/ {print substr($2, 2, length($2) - 2)}' $pkg)"
+  comment=$(awk -F '=' '/pkgdesc/ {print substr($2, 2, length($2) - 2)}' "$pkg")
   echo "$comment"
 }
 
@@ -240,16 +240,16 @@ add_source()
   pkg="$1"
 
   # if no name is provided, add ba-$pkgname to PKGBUILD file
-  if [ -z $2 ]; then
+  if [ -z "$2" ]; then
     name='ba-$pkgname.desktop'
   else
     name="$2"
   fi
 
-  cat "$pkg" |  tr '\n' '\0'\ | \
+  tr '\n' '\0'\ < "$pkg" | \
     sed  "s/\(\x00source=([^)]\+\))/\1\n        \"$name\"\)/" | \
     tr '\0' '\n' > "$pkg.mod"
-  mv "$pkg.mod" $pkg
+  mv "$pkg.mod" "$pkg"
 }
 
 
@@ -262,10 +262,10 @@ add_dep()
 {
   pkg="$1"
 
-  cat "$pkg" |  tr '\n' '\0'\ | \
+  tr '\n' '\0'\ < "$pkg" | \
     sed "s/\(\x00depends=([^)]\+\))/\1 \'blackarch-menus\'\)/" | \
     tr '\0' '\n' > "$pkg.mod"
-  mv "$pkg.mod" $pkg
+  mv "$pkg.mod" "$pkg"
 }
 
 
@@ -280,11 +280,11 @@ add_shasum()
   pkg="$1"
   path="$2"
 
-  sum=$(sha512sum $path | awk '{print $1}')
-  cat "$pkg" |  tr '\n' '\0'\ | \
+  sum=$(sha512sum "$path" | awk '{print $1}')
+  tr '\n' '\0'\ < "$pkg" | \
     sed "s/\(\x00sha512sums=([^)]\+\))/\1\n            \'$sum\'\)/" | \
     tr '\0' '\n' > "$pkg.mod"
-  mv "$pkg.mod" $pkg
+  mv "$pkg.mod" "$pkg"
 }
 
 
@@ -296,8 +296,8 @@ add_shasum()
 incr_release()
 {
   pkg="$1"
-  cat $pkg | awk -F'=' '/pkgrel/{$2=$2+1}1' > "$pkg.mod"
-  mv "$pkg.mod" $pkg
+  awk -F'=' '/pkgrel/{$2=$2+1}1' < "$pkg" > "$pkg.mod"
+  mv "$pkg.mod" "$pkg"
 }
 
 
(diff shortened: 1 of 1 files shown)


ikstream:
To generate the desktop files clone this repo and switch to branch.

Create files for all packages

cd scripts
bash gen-desktop-files -a
Create desktop file for a single package

cd scripts
bash gen-desktop-files -p ../packages/<package>/PKGBUILD 
Create desktop file with icon name different from package name

cd scripts
bash gen-desktop-files -p ../packages/<package>/PKGBUILD  -i <name of icon without png ending>
-a and -p will check if a .png exist in the package directory and add it, if available

ikstream:
Added `-n, --no-version-bump` flag to enable use of the script during new package creation

Edu4rdSHL (changes requested):
Add the suggestions that are suggested in https://www.shellcheck.net/

ikstream:
`-i` option copies the provided file to package dir and rename it to `$packagename.png`

Edu4rdSHL (approved):
Hey @ikstream, awesome! LGTM, I will do some testing and then will merge into master if no issues are found.

ikstream:
Any updates?

Edu4rdSHL:
I have detected the following issue:

When I run the script directly in the PKGBUILD directory, the file is named in an unexpected way. For example:

```
$ cd ~/BlackArch/blackarch/packages/amass
$ bash gen-desktop-files -p PKGBUILD
```

The `.desktop` file is called `ba-..desktop`, maybe we can take the name from the PKGBUILD `pkgname` variable instead of relying on the directory PATH? Also, the `Name` and `Icon` got the `.` value. Here's the resulting file:

```
[Desktop Entry]
Name="."
Icon="."
Comment="In-depth subdomain enumeration written in Go."
TryExec=/usr/bin/Amass
Exec=sh -c '/usr/bin/"Amass";$SHELL'
StartupNotify=true
Terminal=true
Type=Application
Categories=" X-Blackarch-Scanner; X-Blackarch-Recon;"
```

ikstream:
Thanks for the feedback, will look into it and make it work from the PKGBUILD dir

ikstream:
The script uses the `pkgname` of a package now for the `Name` entry in the desktop file.

In addition I added the `-P, --path` flag to provide a path to the packages directory containing the sub-directories with PKGBUILD files. This does only work with -a | --all and -l | --letter flags. 

Example:
To generate desktop files for all `air*` packages from `blackarch` git root run 
```
bash gen-desktop-files -l air -P packages
```

scripts/gen-desktop-files
```diff
@@ -0,0 +1,517 @@
...
+#   pkg: PKGBUILD file
+#   name: desktop/icon file name
+#
+add_source()
+{
+  pkg="$1"
+
+  # if no name is provided, add ba-$pkgname to PKGBUILD file
+  if [ -z "$2" ]; then
+    name='ba-$pkgname.desktop'
```
Edu4rdSHL, line 244: Expressions don't expand in single quotes, use double quotes for that.
```suggestion
    name="ba-$pkgname.desktop"
```

scripts/gen-desktop-files
```diff
@@ -0,0 +1,517 @@
...
+
+# Generate ba-*.desktop files for all packages
+#
+# Arguments:
+#   path - path to PKGBUILD files
+generate_all()
+{
+  path="$1"
+
+  for pkg in $(awk '/groups/ {print FILENAME}' "${path}"/*/PKGBUILD)
```
Edu4rdSHL, line 389:
```suggestion
  awk '/groups/ {print FILENAME}' "${path}"/*/PKGBUILD |  while IFS= read -r pkg
```

Edu4rdSHL (changes requested):

scripts/gen-desktop-files
ikstream: It's the intention, that it is not expanded here, but used in PKGBUILD file

Edu4rdSHL: Ok, please review the other suggestions as well.

scripts/gen-desktop-files
```diff
@@ -0,0 +1,517 @@
...
+#
+# Arguments:
+#   pkg : PKGBUILD file
+#
+write_desktop_file()
+{
+  pkg="$1"
+  name="$(get_name "$pkg")"
+  comment="$(get_desc "$pkg")"
+  ex_name="$(get_prog_name "$pkg")"
```
Edu4rdSHL, line 353:
```suggestion
  pkg="$1"
  name="$(get_prog_name "$pkg")"
  comment="$(get_desc "$pkg")"
  ex_name="$(get_name "$pkg")"
```

scripts/gen-desktop-files
```diff
@@ -0,0 +1,517 @@
...
+      icon="utilities-terminal"
+    else
+      icon="$name"
+    fi
+  else
+    cp "$ICON" "$(dirname "$pkg")/${name}.png"
+    icon="$name"
+  fi
+
+  cat > "$(dirname "$pkg")/ba-${name}.desktop" << EOF
```
Edu4rdSHL, line 366:
```suggestion
  cat > "$(dirname "$pkg")/ba-${ex_name}.desktop" << EOF
```

Edu4rdSHL (changes requested):
Currently, the package binary path is wrong if the PKGBUILD contains a `$_pkgname` as it takes the value from there, the usage of the functions were the other way around. Also, to preserve lowercase `.desktop` filenames we need to point the name to `$ex_name` instead of `$name`

scripts/gen-desktop-files
```diff
@@ -0,0 +1,517 @@
...
+  if [ -z "$ICON" ]; then
+    if [ ! -f "$(dirname "$pkg")/${name}.png" ]; then
+      icon="utilities-terminal"
+    else
+      icon="$name"
+    fi
+  else
+    cp "$ICON" "$(dirname "$pkg")/${name}.png"
+    icon="$name"
+  fi
```
Edu4rdSHL, line 364:
```suggestion
  if [ -z "$ICON" ]; then
    if [ ! -f "$(dirname "$pkg")/${ex_name}.png" ]; then
      icon="utilities-terminal"
    else
      icon="$ex_name"
    fi
  else
    cp "$ICON" "$(dirname "$pkg")/${ex_name}.png"
    icon="$ex_name"
  fi
```

Edu4rdSHL (changes requested):

Edu4rdSHL:
I forgot to mention, we also need to check if the `ba-$pkgname.desktop` path has been added already to the PKGBUILD, otherwise, we end with something like:

```
source=('git+https://github.com/OWASP/Amass.git'
        "ba-$pkgname.desktop"
        "amass.png"
        "ba-$pkgname.desktop"
        "amass.png"
        "ba-$pkgname.desktop"
        "amass.png"
        "ba-$pkgname.desktop"
        "amass.png")
```

ikstream:
The script will now check if there already is a ba-*.desktop file                   
present. If the -f flag is not provided, the script will terminate.                 
                                                                                    
If you want to force an overwrite, because e.g. you want to provide                 
an icon for the application use                                                     
                                                                                    
```                                                                                 
-f, --force                                                                         
```                                                                                 
                                                                                    
this will update existing shasums in the PKGBUILD file                              
                                                                                    
No duplicate entries should be made in the install section or at source

ikstream:
@noraj do you have any comments? Otherwise I would merge the script.

noraj:
> @noraj do you have any comments? Otherwise I would merge the script.

Nop, the idea looks nice. But I don't use a BA menu.

PR #3148, opened 2021-06-29
burpsuite: version bump.

wykwit:
burpsuite version bump and url update so that it points to the latest release

### DIFF
--- a/packages/burpsuite/PKGBUILD
+++ b/packages/burpsuite/PKGBUILD
@@ -2,14 +2,14 @@
 # See COPYING for license details.
 
 pkgname=burpsuite
-pkgver=2021.5.3
+pkgver=2021.6.2
 pkgrel=1
 epoch=1
 groups=('blackarch' 'blackarch-webapp' 'blackarch-proxy' 'blackarch-scanner'
         'blackarch-fuzzer')
 pkgdesc='An integrated platform for attacking web applications (community edition) + SHELLING plugin.'
 arch=('any')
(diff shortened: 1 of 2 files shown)


packages/burpsuite/PKGBUILD
```diff
@@ -2,14 +2,15 @@
...
-pkgver=2021.5.3
+pkgver=2021.6.2
 pkgrel=1
 epoch=1
 groups=('blackarch' 'blackarch-webapp' 'blackarch-proxy' 'blackarch-scanner'
         'blackarch-fuzzer')
 pkgdesc='An integrated platform for attacking web applications (community edition) + SHELLING plugin.'
 arch=('any')
 url='https://portswigger.net/burp/releases/professional-community-2021-2-1?requestededition=community'
+url='https://portswigger.net/burp/releases/community/latest'
```
ikstream, line 13: Why do you add a `url` instead of updating the old one?

Please update `url` instead, as it is outdated anyways

ikstream (changes requested):

packages/burpsuite/PKGBUILD
wykwit: whoops, overlooked it by mistake

noptrix:
@wykwit thank you. we updated it internally. btw, you don't have to touch `source`. simply `pkgver` is enough as long as the url to `source` didn't change. thanks again. closing here

wykwit:
@noptrix, there's still one more thing - the `url` update. Could you commit that as well?

PR #3149, opened 2021-06-29
cloudsploit: fix: Allow arguments to be parsed

Jab2870:
Currently, arguments are not parsed to cloudsploit when run using the script
stored in /usr/bin.

This commit fixes that by parsing each argument that was sent to the script to
cloudsploit


noptrix:
@Jab2870 thank you for this. can you please escape `$` here? e.g.: `"\$@"`

noptrix:
@Jab2870 thank you.

PR #3152, opened 2021-06-29
pyspf : initial commit

D4imyo:
when making the python2 of the library it showed error while installing :
``` 
error: failed to commit transaction (conflicting files)
/usr/bin/spfquery.py exists in both 'python-spf' and 'python2-spf'
```
check the package name variables, this a tool dependency between


noptrix:
@D4imyo thanks for this. watch my changes afterwards please.

PR #3158, opened 2021-07-04
puremagic: initial commit

D4imyo:
a tool dependency.
i am aware of that the library license is MIT and should be installed, but i am confused where to write the install command

### DIFF
--- a/packages/PKGBUILD
+++ b/packages/PKGBUILD
@@ -1,0 +1,46 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=python-puremagic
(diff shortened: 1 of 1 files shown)


noraj:
>  i am aware of that the library license is MIT and should be installed, but i am confused where to write the install command

once in each `package_()`

D4imyo:
so for package_python2-puremagic() :
install -Dm 644 LICENSE "$pkgdir/usr/share/liceenses/python2-puremagic/LICENSE" ?

noraj:
> so for package_python2-puremagic() :
> install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/python2-puremagic/LICENSE" ?

Yes and the same for py3

noptrix:
thank you @D4imyo

PR #3163, opened 2021-07-11
Add redress

wykwit:
https://github.com/goretk/redress

>The redress software is a tool for analyzing stripped Go binaries compiled with the Go compiler. It extracts data from the binary and uses it to reconstruct symbols and performs analysis.

Redress is useful as a standalone binary, but it can also be executed from within radare2. That's when redress truly shines.

>It will extract all the functions and methods and construct symbol flags for them. It will also extract the types in the binary. It will mark the corresponding `_type` with this flag to make analysis easier. By doing this, it highlights which actual type is being allocated on the heap for easier analysis.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,1 +1,2 @@
 github-dorks
+redress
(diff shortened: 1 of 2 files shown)


noptrix:
@wykwit thanks a lot for this!

PR #3182, opened 2021-07-21
pwcrack: update version

noraj:
fix https://github.com/L-codes/pwcrack-framework/issues/7#issuecomment-883847887

### DIFF
--- a/packages/pwcrack/PKGBUILD
+++ b/packages/pwcrack/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=pwcrack
 _pkgname=pwcrack-framework
-pkgver=1.16.1
+pkgver=1.16.2
 pkgrel=1
 groups=('blackarch' 'blackarch-cracker')
 pkgdesc='Password hash automatic cracking framework.'


noptrix:
@noraj when you do a version bump you also need to update the SHA sum using `updpkgsums`. then simply push to repo no need for PR on version bumps.

noptrix:
btw, i just saw that there is no pkgver() and we use 'git'... we need to add pkgver(). i will do this

noraj:
> @noraj when you do a version bump you also need to update the SHA sum using `updpkgsums`. then simply push to repo no need for PR on version bumps.

Yes I forgot to do that.



> btw, i just saw that there is no pkgver() and we use 'git'... we need to add pkgver(). i will do this

pkgver for git usually use git tags but none are available so the fallback is to use commit numbers but since there is a version available https://github.com/L-codes/pwcrack-framework/blob/7581335e45e9612741077aa9761d63aa377b5d58/pwcrack#L7 I find it better to use it. Does a pkgver() for a git source must be using git command or can it use a grep/cut on `pwcrack` script file?

PR #3201, opened 2021-08-03
Update to Mobsf 3.4 

jaykijay:
Some paths changed, quark-engine now needed, Migration will eventually fail because of this i added a message to .install.


noptrix:
@jaykijay thanks a lot for this.

PR #3218, opened 2021-08-16
python-google-cloud-resource-manager: Python tests removed

net-wayfarer:
Fixed the inclusion of Python tests within `package() {}` as it causes file conflicts with other Python packages when installed.


noptrix:
@net-wayfarer thanks:)

PR #3220, opened 2021-08-20
Add slopshell as package?

oldkingcone:
Suggesting to add slopShell.


noptrix:
@oldkingcone please check HOWTO-DEV.md and check `blackarch-pkgbuilds` repository and take one PKGBUILD as an example to work with which would fit to your tool

PR #3232, opened 2021-09-08
proxmark3 : initial commit

### LINKED ISSUE #3191: RRG/Iceman Proxmark3
- **Name**:  RRG/Iceman Proxmark3
- **Source URL**:  https://github.com/rfidresearchgroup/proxmark3
- **Description**:
A general purpose RFID tool for Proxmark3 hardware

There is another popular Proxmark3 repository beside the offical one.

D4imyo:
closes #3191.
i had to add `PREFIX=/usr` because without it the installation process will raise an `permission denied` error while copying binaries.

### DIFF
--- a/packages/proxmark3/PKGBUILD
+++ b/packages/proxmark3/PKGBUILD
@@ -1,0 +1,34 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=proxmark3
+pkgver=v4.13441.r968.gbac58d2a6
+pkgrel=1
+pkgdesc='A general purpose RFID tool for Proxmark3 hardware.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-misc')
+url='https://github.com/rfidresearchgroup/proxmark3'
+license=('GPL2')
+depends=('readline' 'bzip2' 'arm-none-eabi-gcc' 'arm-none-eabi-newlib' 'qt5-base' 'bluez')
+makedepends=('git')
+source=("git+https://github.com/rfidresearchgroup/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+build() {
+ cd $pkgname
+
+ make
+}
+
+package() {
+  cd $pkgname
+
+  make DESTDIR="$pkgdir" PREFIX=/usr install
+}
+


noraj (approved):
LGTM but I didn't test

packages/proxmark3/PKGBUILD
```diff
@@ -0,0 +1,34 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=proxmark3
+pkgver=v4.13441.r968.gbac58d2a6
+pkgrel=1
+pkgdesc='A general purpose RFID tool for Proxmark3 hardware.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-misc')
```
noraj, line 9: Maybe `blackarch-hardware` and/or `blackarch-wireless`. See https://gist.github.com/noraj/5045be32490f1161161e16fc5820066c#ba to list categories. Try avoiding using misc by default but only when no category fits.

noraj (changes requested):

packages/proxmark3/PKGBUILD
noraj: also see https://github.com/BlackArch/blackarch/blob/f27c5c54aaf4dfc1b1d4bbbd41110952dc5d7773/packages/proxmark/PKGBUILD#L7, there is the radio group.

noraj:
By the way what is the difference with https://github.com/Proxmark/proxmark3 that is also proxmark 3 and already packaged as `proxmark` on BA?

noraj:
Builds correctly: `ba-dev -e 'proxmark3 -h' -p proxmark3-v4.13441.r985.gb45602812-1-x86_64.pkg.tar.zst`

D4imyo:
> By the way what is the difference with https://github.com/Proxmark/proxmark3 that is also proxmark 3 and already packaged as `proxmark` on BA?

actually i didn't know that it's already packaged , but i think there's no difference between them.
what do you think ?

D4imyo:
oh i found the differences between them , here it is https://github.com/rfidresearchgroup/proxmark3#what-has-changed

noraj:
> oh i found the differences between them , here it is https://github.com/rfidresearchgroup/proxmark3#what-has-changed

So I guess it's valid to have both, maybe we should rename this one `proxmark3-rdv4`? What do you think @noptrix?

noptrix:
> > oh i found the differences between them , here it is https://github.com/rfidresearchgroup/proxmark3#what-has-changed
> 
> So I guess it's valid to have both, maybe we should rename this one `proxmark3-rdv4`? What do you think @noptrix?

yo, `proxmark3` is the way to go here. thank you guys.

iceman1001:
the official repo  (proxmark/proxmark3)  vs rrg/iceman is quite changed.  While official repo has stagnated,  the latter is alive.

Thank you for adding it.

iceman1001:
There is a package in the Pentoo which would be of assistance.   Since the Pm3 project has a client and a firmware part.

PR #3247, opened 2021-09-28
local-php-security-checker/lpsc : initial commit

### LINKED ISSUE #3244: local-php-security-checker
- **Name**: local-php-security-checker
- **Source URL**: https://github.com/fabpot/local-php-security-checker
- **Description**: PHP security vulnerabilities checker

D4imyo:
closes #3244

note : i named the binary/pkgname `lpsc`

### DIFF
--- a/packages/lpsc/PKGBUILD
+++ b/packages/lpsc/PKGBUILD
@@ -1,0 +1,42 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=lpsc
+_pkgname='local-php-security-checker'
+pkgver=v1.2.0.r0.gf883cc5
+pkgrel=1
+pkgdesc='A command line tool that checks your PHP application packages with known security vulnerabilities.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-scanner')
+url='https://github.com/fabpot/local-php-security-checker'
+license=('AGPL3')
+makedepends=('git' 'go')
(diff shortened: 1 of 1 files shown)


packages/lpsc/PKGBUILD
```diff
@@ -0,0 +1,42 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=lpsc
```
noraj, line 4: Could you please keep `local-php-security-checker` as pkgname but keep `lpsc` as binary name. Or rather keep `local-php-security-checker` as binary name but add `lpsc` as "alias" (not a true shell alias but a symlink) because it's always annoying to not be able to find the binary name without having to read the PKGBUILD for users.

So 

```
pkgname=local-php-security-checker
_pkgname=lscp
```

noraj (changes requested):

noraj:
Thanks @D4imyo for having worked on this!

packages/lpsc/PKGBUILD
D4imyo: how about naming both `pkgname` and the binary  `local-php-security-checker` as users wouldn't even know that `lpsc` existed when they install the package

noraj (approved):

packages/lpsc/PKGBUILD
noraj: Yes in the end people can create their own aliases

noptrix:
@D4imyo thank you very much

PR #3263, opened 2021-11-04
Fix `gitdump` caller script

Kibouo:
The caller script `cd`'d into the install dir, while `git-dump.py` has its output directory hard coded as `outputFolder = "output/"`. So the output would be written to the install dir. This was bad practice, confusing, and required root to run the tool.


noraj (approved):

noptrix:
@Kibouo thank you.

PR #3264, opened 2021-11-06
rustcat: pkgbuild revision

FabioLolix:
~Corrected pkgver() function~
Install license (MIT)
~Removed pointless custom variable~

### DIFF
--- a/packages/rustcat/PKGBUILD
+++ b/packages/rustcat/PKGBUILD
@@ -4,7 +4,7 @@
 pkgname=rustcat
 _binname=rcat
 pkgver=v2.0.0.r6.g95ac0c4
-pkgrel=1
+pkgrel=2
 pkgdesc='A modern port listener and reverse shell.'
 groups=('blackarch' 'blackarch-networking')
 arch=('x86_64' 'aarch64')
--- a/packages/rustcat/PKGBUILD
+++ b/packages/rustcat/PKGBUILD
@@ -5,7 +5,7 @@ pkgname=rustcat
 _binname=rcat
 pkgver=v2.0.0.r6.g95ac0c4
 pkgrel=1
-pkgdesc='A Modern Port Listener & Reverse Shell.'
+pkgdesc='A modern port listener and reverse shell.'
 groups=('blackarch' 'blackarch-networking')
 arch=('x86_64' 'aarch64')
 url='https://github.com/robiot/rustcat'
@@ -30,5 +30,6 @@ package() {
   cd $pkgname
 
   install -Dm 755 "target/release/$_binname" "$pkgdir/usr/bin/$_binname"
+  install -D LICENSE -t "$pkgdir/usr/share/licenses/$pkgname"
 }
 


packages/rustcat/PKGBUILD
```diff
@@ -17,7 +16,7 @@ source=("git+https://github.com/robiot/$pkgname.git")
 pkgver() {
   cd $pkgname
 
-  echo $(git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g')
+  git describe --long --tags | sed 's/^v//;s/\([^-]*-g\)/r\1/;s/-/./g'
```
noraj, line 19: this is not how it works, see our PKGBUILD templates https://github.com/BlackArch/blackarch-pkgbuilds/blob/a8f597f28e9ec58ac8aa300f5825237c2aaf46b8/PKGBUILD-generic-git#L17-L25

packages/rustcat/PKGBUILD
```diff
@@ -29,6 +28,7 @@ build() {
 package() {
   cd $pkgname
 
-  install -Dm 755 "target/release/$_binname" "$pkgdir/usr/bin/$_binname"
+  install -Dm 755 "target/release/rcat" -t "$pkgdir/usr/bin/"
```
noraj, line 31: keep the `_binname` variable

packages/rustcat/PKGBUILD
```diff
@@ -29,6 +28,7 @@ build() {
...
+  install -D LICENSE -t "${pkgdir}/usr/share/licenses/${pkgname}"
```
noraj, line 32: don't use the `${xx}` syntax when unneeded

noraj (changes requested):

packages/rustcat/PKGBUILD
FabioLolix: why?

noraj: Using a variable instead of hardcoding a string in the PKGBUILD is our standard way to do it (https://github.com/BlackArch/blackarch-pkgbuilds/) even if the variable is used only once.

packages/rustcat/PKGBUILD
```diff
@@ -29,6 +28,7 @@ build() {
 package() {
   cd $pkgname
 
-  install -Dm 755 "target/release/$_binname" "$pkgdir/usr/bin/$_binname"
+  install -Dm 755 "target/release/rcat" -t "$pkgdir/usr/bin/"
+  install -D LICENSE -t "$pkgdir/usr/share/licenses/pkgname"
```
noraj, line 32: you have dropped the final `$` before pkgname

noraj (changes requested):

packages/rustcat/PKGBUILD
FabioLolix: yeah, corrected

packages/rustcat/PKGBUILD
```diff
@@ -5,7 +5,7 @@ pkgname=rustcat
 _binname=rcat
 pkgver=v2.0.0.r6.g95ac0c4
 pkgrel=1
```
noraj, line 7: When a package is changed and the `pkgver` is not increased then pkgrel must be increased so `pkgrel=2`

noraj (changes requested):

packages/rustcat/PKGBUILD
FabioLolix: done

noraj (approved):
LGTM

noraj:
@FabioLolix Thanks for your contribution :)

PR #3281, opened 2021-12-05
nettacker: fix typo in group name

wykwit:
Nettacker is the only package in `blackarch-scanners` group. I think it was supposed to be in `blackarch-scanner` instead.


noptrix:
@wykwit thanks a lot.

PR #3301, opened 2022-01-05
theharvester: added python-ujson

nodauf:
python-ujson is a missing dependecie for theharvester


noptrix:
@nodauf thank you! please next time `pkgrel++` or only version bump enough, if available.

PR #3314, opened 2022-01-16
Fix virtualenv dependencies

ikstream:


noptrix:
@ikstream thank you very much. i will update a few things then. please check my changes afterwards.

PR #3315, opened 2022-01-20
python2-virtualenv: move makedepends to depends

ikstream:


noptrix:
@ikstream thx!

PR #3331, opened 2022-02-06
bundle: remove deprecated path option

noraj:
no need to pkgrel++ and release, it can wait next update, it's just to remove teh deprecated flag and avoid the warning


noptrix:
@noraj thank you very much!

PR #3332, opened 2022-02-10
Cutter appimage

dualfade:
please review --


noraj:
It's already packaged here [community/rz-cutter/](https://archlinux.org/packages/community/x86_64/rz-cutter/)

dualfade:
When they update to 2.0.5 will be nice.

PR #3347, opened 2022-03-03
bloodhound fix for 4.0.3 to 4.1.0 upgrade

noraj:
#3345 wasn't enough actually

PR #3348, opened 2022-03-03
upgrade to neo4j 4

noraj:
to come with #3347 I upgraded neo4j

I'm not sure if it's perfect but at least it works


noptrix:
@noraj thanks a lot

noraj:
@noptrix I tested it works well, people that had it previously installed may require to (re)move `/var/lib/neo4j/data/databases/graph.db`.

PR #3352, opened 2022-03-09
autorecon: Fixes current PATH

SecT0uch:
By default autorecon outputs to ./results, which we expect to be the current directory


noraj:
Hi @SecT0uch

Did you tested it works? Because many tools have relative path calls and it may brake the behavior.

SecT0uch:
Yep, running it right know, with expected results.

noraj (approved):

noptrix:
@SecT0uch thank you

PR #3362, opened 2022-03-23
ikeforce: fix deps

noraj:
Fix this issue:

```
$ ikeforce 10.0.0.1
/usr/lib/python2.7/site-packages/OpenSSL/crypto.py:14: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in the next release.
  from cryptography import utils, x509
  File "ikeforce.py", line 13, in <module>
    import ikehandler
  File "/usr/share/ikeforce/ikehandler.py", line 10, in <module>
    import ikecrypto
  File "/usr/share/ikeforce/ikecrypto.py", line 13, in <module>
    from Crypto.Cipher import *
ImportError: No module named Crypto.Cipher
```

### DIFF
--- a/packages/ikeforce/PKGBUILD
+++ b/packages/ikeforce/PKGBUILD
@@ -3,13 +3,13 @@
 
 pkgname=ikeforce
 pkgver=30.575af15
-pkgrel=1
+pkgrel=2
 pkgdesc='A command line IPSEC VPN brute forcing tool for Linux that allows group name/ID enumeration and XAUTH brute forcing capabilities.'
 groups=('blackarch' 'blackarch-cracker')
 arch=('any')
 url='https://github.com/SpiderLabs/ikeforce'
 license=('custom:unknown')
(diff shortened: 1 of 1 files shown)


noraj:
Other topic:

Some actions seems to fails

```
  File "ikeforce.py", line 307, in <module>
    iCookie = ikeneg.secRandom(8).encode('hex')
  File "/usr/share/ikeforce/ikeclient.py", line 46, in secRandom
    randomBytes = OpenSSL.rand.bytes(bytes)
AttributeError: 'module' object has no attribute 'rand
```

Upstream repository is archived, the tool is very old and unmaintained, its using python2, so if it has more issues we should "archive" it. We should have an archive/deprecated or whatsoever category. Tools in this category won't be built or distributed in the repository, they won't be show on the website, etc. but the source PKGBUILD would remain in the repository for archive purpose.

PR #3402, opened 2022-05-04
+ python-boto3 lib

dualfade:
Required dep for s3scanner fix --
New templates used --
please review --

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-boto3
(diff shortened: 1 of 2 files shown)


noptrix:
@dualfade yo, thx for this. `python-boto3` is already present in arch community :) so we don't use to package it on our own. i am closing this. feel free to re-open if something wrong

PR #3403, opened 2022-05-04
+ s3scanner package update; fix

dualfade:
Not working  --

The fix:
repackage s3scanner with pypi template update and build --
using pyproject.toml installation method --
please review --


noptrix:
@dualfade thx for this!

PR #3415, opened 2022-05-12
jackdaw: update dependencies

noraj:
AL version of python-itsdangerous, python-werkzeug, python-flask, python-jinja are not compatible with those asked https://github.com/skelsec/jackdaw/blob/master/setup.py

skelsec is saying me it's a nightmare to maintain and update as the dependencies are changing too fast

so I suggest porting this package to a virtualenv


noptrix:
@noraj r u gonna provide a venv version for us?

PR #3433, opened 2022-05-26
IVRE: update source repos, remove unused dep

p-l-:
This PR:
* (updates the current dev version)
* removes the unneeded dependency to python2-setuptools (Python 2 is no longer supported by IVRE)
* updates the repository URL (the project has moved from cea-sec org to its own org; the redirection works but it's better to use the correct URL anyway)
* adds a `-blackarch-1` to the internal version number (makes debugging easier)


noraj (approved):
LGTM

its0x08 (approved):

PR #3451, opened 2022-06-25
catana: Added catana tool

D3vil0per:
Added PKGBUILD file for catana tool (source: https://github.com/D3vil0per/catana)

PR #3452, opened 2022-06-25
Added catana tool

D3vil0per:


noraj (commented):
Hi see it for #3451. But you should edit this file in the same PR.

D3vil0per:
@noraj one question: since the PKGBUILD takes directly the source by my Repo GitHub, if I change the code of catana file, the package in BlackArch repo is updated automatically or I need to perform some specific step? Because I added some line of code inside that script but when I retrieve it by pacman -Syyu catana, the /usr/bin/catana file is the old one.

ikstream:
You need to adjust the version number in the PKGBUILD file and the package needs to be rebuild

PR #3490, opened 2022-08-18
crackmapexec: package refactoring

noraj:
crackmapexec: package refactoring

- better pkgver
- move from setuptools to PEP517
- remove tons of useless dependencies
- simplify packaging
- remove useless patch
- update upstream repository url

issued faced:

1. `ldapdomaindump` module missing because AL version of impacket is outdated and not requiring it while it should
    - fixed by adding the dependency directly to cme
2. `pypsrp` module missing

For 2.:

while running cme

```
Traceback (most recent call last):                                                                                                                                                                                                           
  File "/usr/bin/cme", line 8, in <module>                                                                                                                                                                                                   
    sys.exit(main())                                                                                                                                                                                                                         
  File "/usr/lib/python3.10/site-packages/cme/crackmapexec.py", line 116, in main                                                                                                                                                            
    args = gen_cli_args()                                                                                                                                                                                                                    
  File "/usr/lib/python3.10/site-packages/cme/cli.py", line 72, in gen_cli_args                                                                                                                                                              
    protocol_object = p_loader.load_protocol(protocols[protocol]['path'])                                                                                                                                                                    
  File "/usr/lib/python3.10/site-packages/cme/loaders/protocol_loader.py", line 12, in load_protocol                                                                                                                                         
    protocol = imp.load_source('protocol', protocol_path)                                                                                                                                                                                    
  File "/usr/lib/python3.10/imp.py", line 170, in load_source                                                                                                                                                                                
    module = _exec(spec, sys.modules[name])                                                                                                                                                                                                  
  File "<frozen importlib._bootstrap>", line 619, in _exec
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/lib/python3.10/site-packages/cme/protocols/winrm.py", line 12, in <module>
    from pypsrp.client import Client
ModuleNotFoundError: No module named 'pypsrp'
```

while installing python-pypsrp

```
WARNING: The directory '/home/noraj/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
Collecting pypsrp==0.7.0
  Downloading pypsrp-0.7.0-py3-none-any.whl (85 kB)
[...]

PR #3491, opened 2022-08-18
[wip] python-pypsrp: refactoring

noraj:
- update dependencies for 0.7
- use fixed version tarball rather than pip in .install file
- PEP517 build system using pyproject.toml
- no risk of dummy package, see #3490

**WARNING** ⚠️ there is still issues to fix for this draft to build


noptrix:
@noraj this is done already.

PR #3502, opened 2022-08-31
add jwt-key-recovery

noraj:


noptrix:
@noraj thanks for this - i will rename `recove.py` to `jwt-key-recover` - we try to use `$pkgname` whenever possible + `recover.py` would most probably conflict/clash with some other program in ArchLInux...

PR #3511, opened 2022-09-11
py lib source refactoring

noraj:
There were 3 options on how to patch `up-libraries`, I choose option n°3 as it was the easiest and requiring the less changes.

```py
                    # option n°1)
                    #   We get a way to retrieve the original URL from the PKGBUILD and propagate it until there, maybe pass source as a second argument to python_packages_version_check(), so that url = old_url, this will be nothing but it's a way to avoid touching to update_pkgbuild()

                    # option n°2)
                    #   set url = None, and path update_pkgbuild() to not update the url when URL=None, but the issue is that even if we don't need to update the URL we need to update the checksum and for that we need ... an URL

                    # option n°3)
                    #   keep url = str(data.get('releases')[available_version][-1]['url']) so that checksum update goes well in update_pkgbuild() but don't update source in the PKGBUILD
```

Squash merge as my commits are dirty.


noptrix:
@noraj <3

PR #3530, opened 2022-09-27
openvas 22.4.0; gsa

dualfade:
please review; This still needs some work.


packages/gsa/PKGBUILD
```diff
@@ -2,41 +2,36 @@
 # See COPYING for license details.
 
 pkgname=gsa
-pkgver=22.4
-pkgrel=3
-pkgdesc='Greenbone Security Assistant.'
-arch=('x86_64' 'aarch64')
-url='https://github.com/greenbone/gsa'
+pkgver=22.4.0
+pkgrel=2
```
noraj, line 6: Else it will be considered as older?

```suggestion
pkgrel=4
```

noraj (changes requested):

PR #3537, opened 2022-09-27
openvas 22.4.0; ospd-openvas

dualfade:
please review; This still needs some work.


dualfade:
This is on the one package having major issues. It will not stay running and is somehow deleting /run/ospd by itself. Also; The unix socket does not fire up for commnuicaiton.

PR #3548, opened 2022-10-08
ghauri initial pr

dualfade:
please review; great new tool under rapid dev.


packages/ghauri/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+depends=('python' 'python-chardet' 'python-tldextract' 'python-colorama'
+         'python-requests')
+makedepends=('git' 'python-setuptools')
+source=("git+https://github.com/r0oth3x49/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd "$pkgname"
+
+  echo "$(git rev-list --count HEAD)"."$(git rev-parse --short HEAD)"
```
noraj, line 21: Use `git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'` please, there are tags available.

noraj (changes requested):

packages/ghauri/PKGBUILD
dualfade: I will adjust. Thanks.

noptrix:
@dualfade we don't need to use quotes in `cd "$pkgname"`, btw. only when you "strcat()" it, e.g.: `"$pkgname-blah"` or `"$pkgname-$pkgver"`, etc. :) i am gonna remove the quotes here.

dualfade:
> @dualfade we don't need to use quotes in `cd "$pkgname"`, btw. only when you "strcat()" it, e.g.: `"$pkgname-blah"` or `"$pkgname-$pkgver"`, etc. :) i am gonna remove the quotes here.

I only did so because the latest shellcheck was insisting ;)  got it. Appreciate it.

PR #3556, opened 2022-10-14
WhisperVPN: Added WhisperVPN tool

D3vil0p3r:
Added PKGBUILD file for WhisperVPN tool (source: https://github.com/WhisperVPN/WhisperVPN)

### DIFF
--- a/packages/WhisperVPN/PKGBUILD
+++ b/packages/WhisperVPN/PKGBUILD
@@ -1,0 +1,27 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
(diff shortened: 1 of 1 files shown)


noraj:
After reviewing the script I'm not sure we would want to add this in our repository. Any thoughts @noptrix?

WhisperVPN:
Hi @noraj I'm the developer behind the WhisperVPN tool. I'd like to know why you don't think that my tools is a good fit for your repository, because i think it will help all of your users in staying safe and anonymous while using any terminal. Lmk :)

noptrix:
@noraj yeah we don't want to add this :)

PR #3559, opened 2022-10-16
python-pypykatz

noraj:
close #3547

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,5 @@
+python-aiosmb
+python-asyauth
+python-msldap
+python-pypykatz
+python-winacl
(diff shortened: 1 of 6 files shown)


SecT0uch:
@noraj What triggers it to be in the repos ? Is there a pipeline to monitor ?
It's been 25 days now

noraj:
You're right it still 0.5.7 in the package repo while it was updated to 0.6.2 on the git repo.

[BlackArch package](https://repology.org/project/python:pypykatz/versions)

In general it's because the package was not added to [lists/to-release](https://github.com/BlackArch/blackarch/blob/master/lists/to-release) which allows to rebuild the package. But it was not the case here. Sometimes it's because the package build fails.

Unfortunately there is no public way to follow that.

@noptrix Could you check if something happened on the build pipeline?

SecT0uch:
If only we had the build logs, we could submit fixes.

I will try to think about [lists/to-release](https://github.com/BlackArch/blackarch/blob/master/lists/to-release) next time.

Thanks :)

SecT0uch:
What's the status on this ?

Should I re-commit to [lists/to-release](https://github.com/BlackArch/blackarch/blob/master/lists/to-release) to trigger the build pipelines again ?

noraj:
In the meantime, you can install it from the git with `blackman -i python-pypykatz`

noraj:
@noptrix Could you take a look?

PR #3563, opened 2022-10-21
brakeman: better package

noraj:
- [x] better pkgver
- [x] fix path issue

## Path issue

When I run `brakeman relative/path/source`, brakeman is looking into `brakeman_install_dir + user_provided_path` (`/opt/brakeman/relative/path/source`) instead of `shell_current_path + user_provided_path` (`~/project/example/relative/path/source`).

```
$ brakeman relative/path/source
Loading scanner...
Please supply the path to a Rails application (looking in /opt/brakeman/relative/path/source).
  Use `--force` to run a scan anyway.
```

Which is due to the `/usr/bin/brakeman` wrapper which do a cd in `/opt/brakeman` but the tool seems to run just fine with a direct `exec`.

So this will avoid to have to do `brakeman "$pwd/relative/path/source"`.

### DIFF
--- a/packages/brakeman/PKGBUILD
+++ b/packages/brakeman/PKGBUILD
@@ -36,8 +36,7 @@ package() {
 
   cat > "$pkgdir/usr/bin/$pkgname" << EOF
 #!/bin/sh
-cd /opt/$pkgname
-exec ./bin/$pkgname "\$@"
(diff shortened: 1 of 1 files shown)


noptrix:
@noraj <3

PR #3567, opened 2022-10-31
#fix python-arsenic: version bump

dualfade:
https://pypi.org/project/arsenic/
please review; version bump and PKGBUILD update.
pkgver=21.8


noptrix:
@dualfade much appreciate

Issue #3572, opened 2022-11-01
Tracking deprecated packages


noraj:
- packerid #2620
- malcom #2857
- rekall #3524
- tyton #3527
- paru #3414 (not that paru is deprecated itself but that it as nothing to do in BA repo so it won't be update, use [AUR](https://aur.archlinux.org/packages/paru) instead)
- gerix-wifi-cracker is broken due to pyqt3 (see #3660) and the upstream repository as disappeared, abandoned since 2011
- inguma use pyqt3 (broken) and python2 (dropped from AL) see #3660, also abandoned since 2008
- qt3, pyqt3 and qscintilla-qt3 # broken, discontinued, not needed anymore, see #3660
- pfff, merged as part of semgrep, see https://github.com/BlackArch/blackarch/issues/2986#issuecomment-1374516413

Also check https://github.com/BlackArch/blackarch/blob/master/lists/broken and https://github.com/BlackArch/blackarch/issues/2759 for broken packages.

noraj:
Everything is tracked here: https://github.com/BlackArch/blackarch/blob/master/lists/broken

PR #3579, opened 2022-11-10
paru: Version bump

ikstream:


noraj (approved):
LGTM but shouldn't we drop paru cf. https://github.com/BlackArch/blackarch/issues/3572?

ikstream:
Why should we drop it. I saw, that there hasn't been an update from our side for a while, but i prefer it over yay.

noraj:
> Why should we drop it. I saw, that there hasn't been an update from our side for a while, but i prefer it over yay.

It's a pacman wrapper not an infosec tool, people can already install it from AUR why package it in BA?

ikstream:
same goes for yay. But it provides a certain ease of usage and doesn't require people to retrieve the PKGBUILD from AUR and build it

PR #3558, opened 2022-10-14
commentor: Added commentor tool

D3vil0p3r:
Added PKGBUILD file for commentor tool (source: https://github.com/D3vil0p3r/comMENTOR)

PR #3596, opened 2022-11-23
Added commentor

D3vil0p3r:


noraj:
You could have added it in #3558, but don't worry I'll care of it

Issue #3593, opened 2022-11-22
Evil-WinRM - OpenSSL::Digest::DigestError happened, message is Digest initialization failed: initialization error


D3vil0p3r:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

When `evil-winrm` tool from BlackArch repo is run, the following output is generated:
```
evil-winrm -u Jareth -p sarah -i 10.10.79.152

Evil-WinRM shell v3.4

Info: Establishing connection to remote endpoint

Error: An error of type OpenSSL::Digest::DigestError happened, message is Digest initialization failed: initialization error

Error: Exiting with code 1
```
Furthermore, each time evil-winrm is run, this git warning is produced:
```
evil-winrm -h            
fatal: not a git repository (or any of the parent directories): .git

<SNIP>
```
This issue occurs also in BlackArch after a Full Upgrade of the system.

In Debian-based OS, Evil-WinRM works correctly.

### Steps to reproduce

- By guessing Arch system has BlackArch repo available, run `sudo pacman -Syyu`
- Run `sudo pacman -Syy evil-winrm`
- Take a test machine where you can test Evil-WinRM, for example this TryHackMe machine: https://tryhackme.com/room/yearoftheowl
- Run `evil-winrm -u Jareth -p sarah -i 10.10.79.152`

where `10.10.79.152` is the Target IP address. It will be different in your case.

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

The following output is produced:
```
Evil-WinRM shell v3.4

Info: Establishing connection to remote endpoint

Error: An error of type OpenSSL::Digest::DigestError happened, message is Digest initialization failed: initialization error

Error: Exiting with code 1
```
Furthermore, each time evil-winrm is run, this git warning is produced:
```
evil-winrm -h            
fatal: not a git repository (or any of the parent directories): .git

<SNIP>
```

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

The following is the expected output:
```
Evil-WinRM shell v3.4

Info: Establishing connection to remote endpoint

*Evil-WinRM* PS C:\Users\Jareth\Documents>
```

### Info for developers

**GNU/Linux distribution:** Arch Linux
**Tool version:** 3.4 (latest)

noraj:
> Run `sudo pacman -Syy evil-winrm`

[...]

PR #3620, opened 2022-12-04
evil-winrm: fix OpenSSL 3 error

D3vil0p3r:
Fixing the issue linked in https://github.com/BlackArch/blackarch/issues/3593 by enabling legacy protocols that Evil-WinRM uses for attacking the target machine. The issue is solved by enabling the legacy protocols after the pkg installation that have been kept disabled by default by OpenSSL maintainers with the latest OpenSSL version.

### DIFF
--- a/packages/evil-winrm/evil-winrm.install
+++ b/packages/evil-winrm/evil-winrm.install
@@ -5,6 +5,8 @@ post_install() {
   bundle config build.nokogiri --use-system-libraries
   bundle config set --local path 'vendor/bundle'
   bundle install
+  sed -zi "s/\[provider_sect\]\ndefault = default_sect/\[provider_sect\]\ndefault = default_sect\nlegacy = legacy_sect/" /etc/ssl/openssl.cnf
+  sed -zi "s/\[default_sect\]\n# activate = 1/\[default_sect\]\nactivate = 1\n\[legacy_sect\]\nactivate = 1/" /etc/ssl/openssl.cnf
 }
 
 post_upgrade() {


packages/evil-winrm/evil-winrm.install
```diff
@@ -5,6 +5,8 @@ post_install() {
   bundle config build.nokogiri --use-system-libraries
   bundle config set --local path 'vendor/bundle'
   bundle install
+  sed -zi "s/\[provider_sect\]\ndefault = default_sect/\[provider_sect\]\ndefault = default_sect\nlegacy = legacy_sect/" /etc/ssl/openssl.cnf
```
noraj, line 8: As said here https://github.com/BlackArch/blackarch/issues/3593#issuecomment-1326805046, we don't want to modify openssl globally / system-wide, we should find a way to create a custom config for evil-winrm and load this one.

noraj (changes requested):

noraj:
You could try adding `openssl-1.1` in dependency but with both 1.1 and 3.0 installed, idk how it will behave.

D3vil0p3r:
> openssl-1.1

I tried to install `openssl-1.1` package, but when I run `evil-winrm` we still get the famous error. So, it still refers to the OpenSSL 3.0 config file.

D3vil0p3r:
> You could try adding `openssl-1.1` in dependency but with both 1.1 and 3.0 installed, idk how it will behave.

I tried to install openssl-1.1 package, but when I run evil-winrm we still get the famous error. So, it still refers to the OpenSSL 3.0 config file.

noraj:
If you take a look at evil-winrm gemspec https://github.com/Hackplayers/evil-winrm/blob/381b126d6a8dc3ca38f7fab73cb80175fcbe8d45/evil-winrm.gemspec#L27, it relies on the [winrm](https://github.com/WinRb/WinRM) gem.

I suggest you open a bug request upstream for openssl 3.0 support.

D3vil0p3r:
> If you take a look at evil-winrm gemspec https://github.com/Hackplayers/evil-winrm/blob/381b126d6a8dc3ca38f7fab73cb80175fcbe8d45/evil-winrm.gemspec#L27, it relies on the [winrm](https://github.com/WinRb/WinRM) gem.
> 
> I suggest you open a bug request upstream for openssl 3.0 support.

Opened to https://github.com/WinRb/WinRM/issues/340

noraj:
By reading https://github.com/WinRb/WinRM/issues/340, I think we can make a workaround.

1. add openssl-1.1 in the dependencies https://github.com/BlackArch/blackarch/blob/ff5e86d11e9d4a7cf731f4f7e16b18d2da166618/packages/evil-winrm/PKGBUILD#L14
2. Modify the wrapper to add an inline `OPENSSL_CONF` in the wrapper `exec OPENSSL_CONF=/path/to/customconf  bundle exec $pkgname.rb "\$@"`https://github.com/BlackArch/blackarch/blob/ff5e86d11e9d4a7cf731f4f7e16b18d2da166618/packages/evil-winrm/PKGBUILD#L42-L46 and ship a custom conf in `/usr/share/evil-winrm`.

PR #3628, opened 2022-12-08
evil-winrm: try workaround idea

noraj:
cf. https://github.com/BlackArch/blackarch/pull/3620#issuecomment-1339237700

### DIFF
--- a/packages/evil-winrm/PKGBUILD
+++ b/packages/evil-winrm/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=evil-winrm
 pkgver=v3.4.r0.g381b126
-pkgrel=3
+pkgrel=4
 epoch=1
 pkgdesc='The ultimate WinRM shell for hacking/pentesting.'
 arch=('any')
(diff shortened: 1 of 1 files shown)


noraj:
@D3vil0p3r Could you try this please?

D3vil0p3r:
> @D3vil0p3r Could you try this please?

Of course yes, I need just some minutes/hour for finishing some activities and I will back with that try

D3vil0p3r:
Hey @noraj sorry for being late.

According to your suggestions, I installed `evil-winrm` from BlackArch repo and as test I created the `/usr/bin/evil-winrm` file as:
```
#!/bin/sh
cd /usr/share/evil-winrm
OPENSSL_CONF=/etc/ssl/evilssl.cnf bundle exec evil-winrm.rb "$@"
```
I removed the `exec` at the beginning of the last line from your suggestion because triggered the error:
```
/usr/bin/evil-winrm: line 3: /usr/share/evil-winrm/OPENSSL_CONF=/etc/ssl/evilssl.cnf: No such file or directory
```
By the way, this editing on the wrapper solved our issue. Indeed by linking it to `evilssl.cnf` (that is the custom OpenSSL config file), it works. By linking to the original OpenSSL config of course file does not.

image

So, this solution works without installing `openssl-1.1`. So, I guess it is possible to solve by editing the wrapper in the manner above inside `PKGBUILD`.

A last note: I see that with this package, each time Evil WinRM is run, there is the following not blocking warning (as you can see from the image above too):
```
fatal: detected dubious ownership in repository at '/usr/share/evil-winrm'
To add an exception for this directory, call:

	git config --global --add safe.directory /usr/share/evil-winrm
```
Is it possible to fix it?

Thank you for your approach.

noraj:
Maybe for the git warning but I PRed upstream to remove the use of git so let's see if they merge it.

D3vil0p3r:
Hello @noptrix, is it possible to merge this PR for fixing the OpenSSL issue that Evil-WinRM suffered all this period?

noptrix:
thank you guys

PR #3629, opened 2022-12-10
ida-free: add package

D3vil0p3r:
PKGBUILD from https://aur.archlinux.org/packages/ida-free with fix to errors during the building.

IDA is a light but powerful tool can quickly analyze the binary code samples and users can save and look closer at the analysis results.


noptrix:
@D3vil0p3r thank you.

Issue #3296, opened 2021-12-21
APKID Yara rules not compatible


jaykijay:
### Bug description

APKID not working after latest update because Yara rules seem to be incompatible

### Steps to reproduce

Install apkid
Run it

**Actual result:**

```
[+] APKiD 2.1.1 :: from RedNaga :: rednaga.io
Traceback (most recent call last):
  File "/usr/bin/apkid", line 33, in <module>
    sys.exit(load_entry_point('apkid==2.1.1', 'console_scripts', 'apkid')())
  File "/usr/lib/python3.10/site-packages/apkid/main.py", line 90, in main
    rules = options.rules_manager.load()
  File "/usr/lib/python3.10/site-packages/apkid/rules.py", line 46, in load
    self.rules = yara.load(self.rules_path)
yara.Error: rules file "/usr/lib/python3.10/site-packages/apkid/rules/rules.yarc" is incompatible with this version of YARA
```

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)
Should work? :D

### Info for developers

**GNU/Linux distribution:** Manjaro (unstable) + Blackarch Repos
**Tool version:** APKID 2.1.1, yara-rednaga

vichhika:
> ### Bug description
> APKID not working after latest update because Yara rules seem to be incompatible
> 
> ### Steps to reproduce
> Install apkid Run it
> 
> **Actual result:**
> 
> ```
> [+] APKiD 2.1.1 :: from RedNaga :: rednaga.io
> Traceback (most recent call last):
>   File "/usr/bin/apkid", line 33, in <module>
>     sys.exit(load_entry_point('apkid==2.1.1', 'console_scripts', 'apkid')())
>   File "/usr/lib/python3.10/site-packages/apkid/main.py", line 90, in main
>     rules = options.rules_manager.load()
>   File "/usr/lib/python3.10/site-packages/apkid/rules.py", line 46, in load
>     self.rules = yara.load(self.rules_path)
> yara.Error: rules file "/usr/lib/python3.10/site-packages/apkid/rules/rules.yarc" is incompatible with this version of YARA
> ```
> 
> **Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour) Should work? :D
> 
> ### Info for developers
> **GNU/Linux distribution:** Manjaro (unstable) + Blackarch Repos **Tool version:** APKID 2.1.1, yara-rednaga
> 
> ### Link to debug log

[...]

PR #3630, opened 2022-12-10
Add python-yara-dex PKGBUILD and replace python-yara-rednaga in PKGBU…

AkechiShiro:
…ILD of apkid

Fix for #3296


packages/apkid/.SRCINFO
```diff
@@ -0,0 +1,19 @@
+pkgbase = apkid
```
noraj, line 1: `.SRCINFO`is use only for the AUR, we don't need that on BA repo.

packages/python-yara-dex/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
```
noraj, line 1: You can find our PKGBUILD templates here : https://github.com/BlackArch/blackarch-pkgbuilds

packages/python-yara-dex/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+# See COPYING for license details.
+
+pkgname=python-yara-dex
+_pkgname=yara-python
```
noraj, line 5: Is that not possible to use [this template](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib-setuptools) instead of what you did manually? It seems it is available on pypi.

packages/python-yara-dex/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgver=1.0.3
+pkgrel=1
+pkgdesc='The Python interface for YARA.'
+url='https://github.com/MobSF/yara-python-dex'
+arch=('any')
+groups=('blackarch')
```
noraj, line 11: We don't add groups for package that are dependencies.

packages/python-yara-dex/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+license=('Apache')
+makedepends=('python-setuptools' 'git' 'glibc' 'python' 'yara')
+provides=('python-yara')
+conflicts=('python-yara' 'python-yara-rednaga')
```
noraj, line 15: If they really conflicts (cf. https://wiki.archlinux.org/title/PKGBUILD#conflicts) have you check that other BA packages are using one of those dependencies? And if it's the case if they can move to `python-yara-dex`, else it will be problematic to have conflicting dependencies.

noraj (changes requested):

packages/apkid/.SRCINFO
AkechiShiro: Oh sorry I thought they would be needed regardless, will delete it.

packages/python-yara-dex/PKGBUILD
AkechiShiro: They do conflict, yes, since APKID needs files from one of them, I haven't tried installing both to try and see if conflicting files are shown and pacman fails, but I believe that's what would happen. 

They could move to `python-yara-dex` but testing case by case would be needed, I have not tried to switch all packages that use `python-yara-rednaga` to `python-yara-dex` but according to this issue/comment : https://github.com/rednaga/APKiD/issues/318#issuecomment-1275072279
 
`python-yara-rednaga` is deprecated now, hence could explain the bug that was encountered with APKID.

But for that, another PR is needed, and I believe also that an issue should be created tracking packages that have moved on the new dependency, so that we can track the status in regards to all the pkgs that depends on `python-yara-rednaga` or `python-yara-1`

packages/python-yara-dex/PKGBUILD
AkechiShiro: I'm having issues figuring out the proper URL for the pypi source, see : 
image
This doesn't seem to exist : https://files.pythonhosted.org/packages/source/y/yara-python-dex/yara-python-dex-1.0.3.tar.gz
Even if it exists here : https://pypi.org/project/yara-python-dex/

If you can help me see, what is the one thing I'm doing wrong?
The source is defined as:
```bash
https://files.pythonhosted.org/packages/source/${_pkgname::1}/$_pkgname/$_pkgname-$pkgver.tar.gz
```
I've also tried : 
```sh
https://files.pythonhosted.org/packages/py3/${_name::1}/$_pkgname/${_pkgname//-/_}-$pkgver-py3-none-any.whl
```

AkechiShiro: So what should we do ? Do I keep the git sources ? What should I do  regarding the source URL?

packages/apkid/.SRCINFO
noraj: You removed one but there is still `packages/apkid/.SRCINFO`

packages/python-yara-dex/PKGBUILD
noraj: As it's not on pypi, the git tags from the git repo could be used instead: https://github.com/MobSF/yara-python-dex/archive/refs/tags/1.0.3.tar.gz

It's just that the auto-update won't work.

packages/python-yara-dex/PKGBUILD
noraj: What is the output of `grep -r python-yara packages`?

packages/python-yara-dex/PKGBUILD
AkechiShiro: I'm running into an issue, I'm not sure why : 
```
  File "/home/username/project/opensource/blackarch/packages/python-yara-dex/src/yara-python-dex-1.0.3/setup.py", line 310, in <module>
    with open('yara-python/README.rst', 'r', 'utf-8') as f:
  File "/usr/lib/python3.10/codecs.py", line 905, in open
    file = builtins.open(filename, mode, buffering)
FileNotFoundError: [Errno 2] No such file or directory: 'yara-python/README.rst'
==> ERROR: A failure occurred in build().
```
Here is the tarball contents, it seems the `setup.py` code tries to read a file that will never be at the right place?
image
I'm pushing along the changes, I've made, so you can take a look, but I'd like your input on what you would recommend, or maybe I'm doing something wrong (using `git+https://` worked fine, but I believe that using the release tag, it should work all the same).

Oh, I got it, it's because it's using a git submodule : 
image and the tarball does not contain the git submodule's files, so should I get the latest release tag of the git submodule as well ?

packages/python-yara-dex/PKGBUILD
AkechiShiro: Here is the output : 
```
packages/yeti/PKGBUILD:         'python-wtforms' 'python-xlsxwriter' 'python-yara' 'mongodb' 'redis'
packages/python-yara-dex/PKGBUILD:pkgname=python-yara-dex
packages/python-yara-dex/PKGBUILD:provides=('python-yara')
packages/python-yara-dex/PKGBUILD:conflicts=('python-yara' 'python-yara-rednaga')
packages/ssma/PKGBUILD:depends=('python' 'python-pefile' 'python-magic' 'python-yara-rednaga' 'ssdeep'
packages/quickscope/PKGBUILD:         'python-pefile' 'quark-engine' 'pyaxmlparser' 'python-yara'
packages/mobsf/PKGBUILD:         'python-requests' 'jre-openjdk' 'python-yara' 'sqlite' 'fontconfig'
packages/mobsf/pkg/mobsf/.BUILDINFO:installed = python-yara-dex-279.32b6a74-4-any
packages/mobsf/pkg/mobsf/.PKGINFO:depend = python-yara
packages/apkid/pkg/apkid/.BUILDINFO:installed = python-yara-dex-279.32b6a74-4-any
packages/apkid/pkg/apkid/.PKGINFO:depend = python-yara-dex
packages/apkid/pkg/apkid/.PKGINFO:makedepend = python-yara-dex
packages/apkid/PKGBUILD:depends=('python' 'python-argparse' 'python-yara-dex')
packages/apkid/PKGBUILD:makedepends=('python-setuptools' 'python-yara-dex')
packages/torcrawl/PKGBUILD:         'python-socks' 'python-urllib3' 'python-yara' 'tor')
packages/oledump/PKGBUILD:depends=('python' 'python-lxml' 'python-yara' 'python-olefile')
packages/pepper/PKGBUILD:depends=('python' 'python-pefile' 'python-yara' 'python-requests'
packages/python-plaso/PKGBUILD:         'python-xlsxwriter' 'python-yara' 'python-cryptography')
packages/peframe/PKGBUILD:         'python-yara' 'python-m2crypto' 'python-virustotal-api' 'python-magic'
packages/python-yara-rednaga/PKGBUILD:pkgname=python-yara-rednaga
packages/python-yara-rednaga/PKGBUILD:provides=('python-yara')
packages/python-yara-rednaga/PKGBUILD:conflicts=('python-yara')
packages/mkyara/PKGBUILD:depends=('python' 'python-capstone' 'python-yara' 'python-argparse')
packages/python-yaraprocessor/PKGBUILD:pkgbase=python-yaraprocessor
packages/python-yaraprocessor/PKGBUILD:pkgname=('python2-yaraprocessor' 'python-yaraprocessor')
packages/python-yaraprocessor/PKGBUILD:package_python-yaraprocessor() {
packages/stacs/PKGBUILD:depends=('python' 'python-click' 'python-pydantic' 'python-yara'
```

packages/python-yara-dex/PKGBUILD
noraj: Ah yeah, git sub-modules are a pain. As there is no source archive on pypi and the tarball on github release has not `.git` that is required for git sub-modules, then the best way would be to do a git style PKGBUILD like https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git

noraj: I'm sorry it's a pain, you didn't pick the easiest for a first contribution but I'm glad continue working on it. 😉

packages/python-yara-dex/PKGBUILD
noraj: - PKGBUILD [yara](https://github.com/archlinux/svntogit-community/blob/packages/yara/trunk/PKGBUILD) = not on pypi = github https://github.com/VirusTotal/yara
- PKGBUILD [python-yara](https://github.com/archlinux/svntogit-community/blob/packages/python-yara/trunk/PKGBUILD) = pypi [yara-python](https://pypi.org/project/yara-python/) = github https://github.com/VirusTotal/yara-python
- PKGBUILD [python-yara-rednaga](https://github.com/BlackArch/blackarch/blob/master/packages/python-yara-rednaga/PKGBUILD) = not on pypi  = github https://github.com/rednaga/yara-python-1
- PKGBUILD python-yara-dex = pypi [yara-python-dex](https://pypi.org/project/yara-python-dex) = github https://github.com/MobSF/yara-python-dex

1. To me it doesn't seem like there are conflicting so you should `provides` and `conflicts`. There are only `python-yara` and `python-yara-rednaga` that conflict each other but `python-yara-rednaga` is already flagged as so. I don't see `python-yara-dex` conflicting with neither of them.
2. To respect convention, this package should be named `python-yara-python-dex` and not `python-yara-dex`. I understand the suggested naming since the PKGBUILD `python-yara` should be named `python-yara-python`. It's common for python packages that include `python` in the name (eg. https://archlinux.org/packages/?q=python-python) to have several times the word `python` in the PKGBUILD name since the convention is `python-<python_package_name>`.

AkechiShiro: Sorry about the delay for any updates.

About 2. : I understand, I've made the change, I'll push it tonight.

In respect to 1 :
So the pkg `python-yara-python-dex` should conflict only with itself ? And provides only `python-yara-python-dex` ? Or it should just conflict with `python-yara-rednaga` and provide it ?

packages/python-yara-dex/PKGBUILD
AkechiShiro: I got around to fix it, just recently, if you could review the changes, thanks !

packages/python-yara-dex/PKGBUILD
noraj: Sorry My sentence was missing the word `remove` . But thinking now, I just tested there was no conflict at the package installation but I didn't check for conflict at namespace import in python. So you should have to take a look at that.

AkechiShiro: I've tried installing both packages `python-yara-rednaga`, `python-yara-python-dex` into a clean chroot and see, if they do collide, and there is a lib that seems to conflict : 
```
python-yara-rednaga: /usr/lib/python3.10/site-packages/yara.cpython-310-x86_64-linux-gnu.so exists in filesystem (owned by python-yara-python-dex)
```
Deleting this file, it seems that it is the only library/file that conflicts for both the packages.

As for the namespace Python, I'm not sure how I'd go around testing if there is indeed a conflict, if you have any advices on how to do that.

Looking into a Python REPL : 
```
>>> import yara
>>> yara.__file__   
'/usr/lib/python3.10/site-packages/yara.cpython-310-x86_64-linux-gnu.so'
```
This shows, that importing the `yara` module uses the library that both the packages share, I don't know if this could break any software, since the libraries are different.

AkechiShiro: I believe then I'm done with this contribution, it was quite lengthy, if there is anything you believe I should improve on or better explain, then feel free to let me know, so this can be merged to fix, the issue with APKID.

noraj: I was wrong to tell you to remove `provides` and `conflicts` as they are really conflicting as you demonstrated.
I don't think that we should move the dependencies of other packages to `python-yara-python-dex` as it will probably generate more issues.
Yes I'm sorry this review is length but it is a complex case and the outcome may impact many packages.
We have two choices here:

1. Keep apkid PKGBUILD as it is and use python-yara-python-dex but with the downside of having conflicting package so mobsf and the other packages you listed here https://github.com/BlackArch/blackarch/pull/3630#discussion_r1051488264 and that we try to avoid.
2. Totally change apkid PKGBUILD structure to use a virtual environment to avoid this dependency nightmare but also at a cost we also try to avoid because it makes much bigger package, more complex and harder to maintain.

I'm not sure myself of what the better options between those two. Any advice @noptrix?

AkechiShiro: I understand, that it is a tricky case to solve due to the multiples yara packages variant, it seems to be a complex, I would lean towards solution 2 in the long run. 

However, I don't understand what you mean exactly by using a virtual env for APKID PKGBUILD, you mean that inside the PKGBUILD : 
1. A Python venv (virtual env) should be built in order to get `python-yara-python-dex` and build APKID using that, then install it.
2. Would APKID PKGBUILD install a virtual env (venv) and keep it for executing itself only with the specific dependency it needs ?

But I do agree that 2 would throw the complexity of maintaining that on the maintainers.

noraj: APKID PKGBUILD using a standard jointscript PKGBUILD like https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-standalone with a wrapper launching a virtual env. and The real whole installation happening in an `.install` script in the `post_install()` hook like that https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/python-standalone.install.

AkechiShiro: Hey, I've tried to do some modification, but I'm still not sure if I understood correctly what is the PKGBUILD python-standalone should contain along with the `.install`.

If I understand correctly, the PKGBUILD just builds a Python standalone virtual env and the `.install` called in post_install will install the dependency inside the virtual env using `pip` ?

But the way the PKGBUILD-python-standalone script example is made, it looks like it's for a python script for instance, here what we're packaging is a library for MobSF : 
image

noraj: > If I understand correctly, the PKGBUILD just builds a Python standalone virtual env and the `.install` called in `post_install()` will install the dependency inside the virtual env using `pip` ?

Yes that's it.

apkid can be used as a CLI tool so this standalone setup would be right for that case. But in the case of apkid is used as a library (dependency) then it can't be installed in a standalone way. There are only two packages depending on apkid: mobsf  and quickscope. [Mobsf](https://github.com/BlackArch/blackarch/tree/master/packages/mobsf) is quick a mess too so a standalone packaging could apply to it too. Then if there is only quickscope left we could install it that way too.

AkechiShiro: This looks like way too much work in order to just fix MobSF, I would rather use it with docker then, I might come back to try and fix this but fixing all this mess is way more than I can handle.
Thanks for your help in trying to move this PR towards a merge but there seems to be many PRs needed to have everything working "cleanly", I'd say

noraj: Yes, most web platform tools are not easily packagable and often dependencies are not very well maintained so it's a nightmare to package.

noraj: Thanks for the efforts.

noraj:
I'll just let it open at the moment as a reminder for me to do something about mobsf and apkid.

noraj:
Note to myself: drop apkid package and package mobsf as standalone.

D3vil0p3r:
> Note to myself: drop apkid package and package mobsf as standalone.

For which reason apkid should be dropped? If so, https://github.com/BlackArch/blackarch/issues/3296 can be closed

noraj:
> For which reason apkid should be dropped?

I don't remember but was probably to hard to package or a dependency hell with too many dependencies. It would be better to have a standalone package for mobsf.

D3vil0p3r:
@AkechiShiro why it is still in draft?

AkechiShiro:
@D3vil0p3r  See https://github.com/BlackArch/blackarch/pull/3630#discussion_r1170427980

D3vil0p3r:
I tested mobsf, I will open a new PR for by using pyproject.toml for install. It seems to work correctly.

What I invite you to do is to open this PR and waiting for merging due to the following statements:
* mobsf uses apkid as dependency (look [here](https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/3cf21cec166cd080deebacc634000595af77c8e6/pyproject.toml#L46)
* apkid latest version uses `python-yara-dex` as dependency (look [here](https://github.com/rednaga/APKiD/blob/4c357088ce88c1027f32b691667f6f55e52f77df/setup.py#L50)
* In BA there is only one more tool dependant on `python-yara-rednaga` that is [ssma](https://github.com/BlackArch/blackarch/blob/master/packages/ssma/PKGBUILD). I tested it by `python-yara-python-dex` dependency and it works correctly (despite the project is unmaintained for years).

According to this, @noraj would you approve the replacement of `yara-rednaga` as dependency of `apkid` and `ssma` by `yara-dex`?

D3vil0p3r:
Once this PR will be merged, I will open this https://github.com/BlackArch/blackarch/pull/4352 for MobSF.

noraj:
> According to this, @noraj would you approve the replacement of `yara-rednaga` as dependency of `apkid` and `ssma` by `yara-dex`?

The plan was:

> Note to myself: drop apkid package and package mobsf (and quicscope) as standalone.

D3vil0p3r:
> The plan was:
> 
> > Note to myself: drop apkid package and package mobsf (and quicscope) as standalone.

`apkid` is still a direct dependency of `mobsf` (evid. [here](https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/3cf21cec166cd080deebacc634000595af77c8e6/pyproject.toml#L46)). If we drop `apkid` how can we be sure mobsf will work correctly?

noraj:
> If we drop `apkid` how can we be sure mobsf will work correctly?

By "package mobsf as standalone" I mean using this [PKGBUILD](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-standalone) and [.install](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/python-standalone.install) and so use a virtual environment.

> If we use PKGBUILD standalone, it leverages on `requirements.txt` file that mobsf does not have anymore.

Not necessarily, it can also be done either from pypi or from pyproject.toml, see:

https://github.com/BlackArch/blackarch-pkgbuilds/blob/b8fd779ec987a8d29f543a5cc5b729074ece638a/python-standalone.install#L9-L14

D3vil0p3r:
Ok, thanks. In old package version, `.install` file has:
```
  echo '==> Clean Up'
  bash scripts/clean.sh y
  echo '==> Migrating Database if exists'
  python manage.py makemigrations
  python manage.py makemigrations StaticAnalyzer
  python manage.py migrate
  echo '==> Installation Complete'
  echo '==> Remove old Database in ~/.MobSF/ if migration failed!'
```
This code is run as root during package installation and it creates `.MobSF` dir in `/root` directory. So, it means that, by keeping this approach, `mbsof` can be run only by `sudo` in order to login on the web application, unless the user runs those commands manually after the install.

D3vil0p3r:
Btw, here the PR https://github.com/BlackArch/blackarch/pull/4352

D3vil0p3r:
`mobsf` and `quickscope` PR has been merged.

D3vil0p3r:
@AkechiShiro to not bother you, I created a new updated PR related to this: https://github.com/BlackArch/blackarch/pull/4377 Feel free to close this PR. We can discuss it on the proper issue ticket https://github.com/BlackArch/blackarch/issues/3296 or new PR itself.

PR #3642, opened 2022-12-20
Deleted MimeType=application/x-executable

D3vil0p3r:
Deleted `MimeType=application/x-executable` because it forces AppImage files to open NetworkMiner when the AppImage has in Properties "Executable as a Program" disabled and when the user double click on AppImage file itself.


DF001:
I can confirm that this issue is happening and it got fixed when i tried the proposed pull request

noptrix:
@D3vil0p3r thank you

PR #3644, opened 2022-12-23
lbd: Change gnu-netcat dependency to openbsd-netcat and adding missing dependency

D3vil0p3r:
gnu-netcat does not provide any output when run as listener while openbsd-netcat does (i.e. Listening on port 4444...) and it makes the user aware that netcat is listening. gnu-netcat and openbsd-netcat cannot be installed together because in conflict but openbsd-netcat is better (used also on other pentesting distros).


D3vil0p3r:
Tested with `openbsd-netcat` package. `lbd` is compatible with `openbsd-netcat`. Both of netcat packages provide the same output, and by inspecting `lbd` code, the `nc` arguments are compliant with both `openbsd` and `gnu` netcat packages.

In particular, the netcat commands called by `lbd` are like `nc $DOMAIN 80`.

D3vil0p3r:
Hello @noptrix did you get the chance to review these changes?

packages/lbd/PKGBUILD
```diff
@@ -3,18 +3,17 @@
...
 pkgver=20130719
-pkgrel=5
+pkgrel=6
 groups=('blackarch' 'blackarch-recon')
 pkgdesc='Load Balancing detector,'
 url='http://ge.mine.nu/code/lbd'
 arch=('any')
 license=('GPL')
-depends=('bash' 'gnu-netcat')
+depends=('bash' 'dnsutils' 'openbsd-netcat')
```
noraj, line 12: Why is `dnsutils` required?

D3vil0p3r: @noraj Because lbd needs `host` command for working. `host` is provided by `dnsutils`.

packages/lbd/PKGBUILD
```diff
@@ -3,18 +3,17 @@
...
 #source=('http://ge.mine.nu/code/lbd')
 source=("$pkgname")
 sha512sums=('2436f9e372af7281a1e3dae1ab9d020933d35277664783fbc39736630a8c8d8acbecb2b5028a593e465ca4729a990daacdd86191570ccdaab912a8021a61a21f')
 
 package() {
   install -Dm 755 $pkgname "$pkgdir/usr/bin/$pkgname"
 }
-
```
noraj, line 20: Beware not to remove the ending blank line.

D3vil0p3r: Ok

PR #3645, opened 2022-12-23
peass: Changed gnu-netcat dependency to openbsd-netcat

D3vil0p3r:
gnu-netcat does not provide any output when run as listener while openbsd-netcat does (i.e. Listening on port 4444...) and it makes the user aware that netcat is listening. gnu-netcat and openbsd-netcat cannot be installed together because in conflict but openbsd-netcat is better (used also on other pentesting distros).

### DIFF
--- a/packages/peass/PKGBUILD
+++ b/packages/peass/PKGBUILD
@@ -11,7 +11,7 @@ url='https://github.com/carlospolop/privilege-escalation-awesome-scripts-suite'
 license=('MIT')
 depends=('bash')
 makedepends=('git')
-optdepends=('fping' 'nmap' 'ping' 'gnu-netcat')
+optdepends=('fping' 'nmap' 'ping' 'openbsd-netcat')
 source=("$pkgname::git+https://github.com/carlospolop/privilege-escalation-awesome-scripts-suite.git")
 sha512sums=('SKIP')
 


noraj:
They don't have the same options supported so we need to check that peass doesn't ask for any option supported in GNU version but not in OpenBSD one.

D3vil0p3r:
@noraj I went deep to it. PEASS is compatible with `openbsd-netcat`. Inside `linpeas_base.sh` there is a piece of code managing  the main netcat packages installed in the system (as also openbsd type):

> https://github.com/carlospolop/PEASS-ng/blob/e29c9e88d5dc5c33682fd2aeb36f3cae60d3fa21/linPEAS/builder/linpeas_parts/linpeas_base.sh#L595-L609

Another piece of code related to netcat in LinPEAS is:

> https://github.com/carlospolop/PEASS-ng/blob/a4b226c16e19e82c73b9a248713fd2b25396d50c/linPEAS/builder/linpeas_parts/linpeas_base.sh#L808-L816

And, according to the differences between openbsd-netcat and gnu-netcat related to the usage of some arguments does not impact LinPEAS.

I tested on an Arch Linux system, and both of netcat packages provide the same output result.

D3vil0p3r:
Closing because the PR with the right changes has been merged: https://github.com/BlackArch/blackarch/pull/3651

PR #3654, opened 2023-01-02
osintgram: add package

D3vil0p3r:
OSINT tool on Instagram. It offers an interactive shell to perform analysis on Instagram account of any users by its nickname.

Source: https://github.com/Datalux/Osintgram


packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+license=('GPL3')
+depends=()
+makedepends=('git' 'python')
+source=("git+https://github.com/Datalux/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
```
noraj, line 20: As explained in [PKGBUILD-generic-git](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git#L23-L24), if [tag exists](https://github.com/Datalux/Osintgram/tags), prefer

```
git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
```

Don't forget to change `pkgver` then.

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,49 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=osintgram
+pkgver=348.50d44fd
+pkgrel=1
+pkgdesc='OSINT tool on Instagram. It offers an interactive shell to perform analysis on Instagram account of any users by its nickname.'
```
noraj, line 7:
```suggestion
pkgdesc='OSINT tool offering an interactive shell to perform analysis on Instagram account of any users by its nickname.'
```

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+arch=('any')
+groups=('blackarch' 'blackarch-recon')
+url='https://github.com/Datalux/Osintgram'
+license=('GPL3')
+depends=()
```
noraj, line 12: You forgot to add runtime dependencies.

https://github.com/Datalux/Osintgram/blob/master/requirements.txt

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
+
+prepare() {
+  cd $pkgname
+
+  python -m venv venv
```
noraj, line 26: don't use a virtual environment, we use those only for very complex cases

all the dependencies in `requirements.txt` are available in AL or BA repo outside of `instagram-private-api` that need to be packaged as well in your PR.

noraj (changes requested):

packages/osintgram/PKGBUILD
D3vil0p3r: @noraj should they not be managed by `pip install -r requirements.txt` in `prepare()`?

packages/osintgram/PKGBUILD
D3vil0p3r: Fixed

packages/osintgram/PKGBUILD
noraj: Only for virtualenv installs that we absolutely try to avoid. Here we don't need a venv nor a `prepare()`. As the next comment state list all the dependencies that are already existing here and package the missing oens (`instagram-private-api`).

packages/osintgram/PKGBUILD
D3vil0p3r: `prepare()` deleted.

packages/osintgram/PKGBUILD
D3vil0p3r: Ok thanks @noraj . Currently I retrieved the `instagram-private-api` source file from https://pypi.org/project/instagram-private-api/#files
I need only to understand how to manage it in this package. I will try to do it tomorrow.

noraj: All you have to do in this case for this lib is filling https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib-setuptools. It should be very easy as there is no deps https://github.com/ping/instagram_private_api/blob/0fda0369ac02bc03d56f5f3f99bc9de2cc25ffaa/setup.py#L34

D3vil0p3r: Thanks. A theorical question: can you provide me the reason why virtualenv installs should be avoided? What are the concerns?

D3vil0p3r: > All you have to do in this case for this lib is filling https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib-setuptools. It should be very easy as there is no deps https://github.com/ping/instagram_private_api/blob/0fda0369ac02bc03d56f5f3f99bc9de2cc25ffaa/setup.py#L34

Do you mean that we need to create a PKGBUILD, so a BA package, for `instagram-private-api`, right? If so, in which folder I should submit it? I don't think in `BlackArch/blackarch/packages` folder.

D3vil0p3r: PS: created PR for `python-instagram-private-api`: https://github.com/BlackArch/blackarch/pull/3657

noraj: You know when you edit a file with github webui to prepare your PR it basically just create a branch `patch-X` on your fork (eg. https://github.com/D3vil0p3r/blackarch/tree/patch-8 here) based on [BlackArch:master](https://github.com/BlackArch/blackarch). So for example here, you can browse [D3vil0p3r:patch-8](https://github.com/D3vil0p3r/blackarch/tree/patch-8) and create `python-instagram-private-api` PKGBUILD there so it will be added to the same PR rather than creating a new PR for it. 1 PR per tool is nice but when it's the dependencies of the tool it's best to have them in the same PR.

D3vil0p3r: Usually, I work by using `git` and on GitHub WebUI I didn't know this stuff. So, if I add something on an already existing fork (like https://github.com/D3vil0p3r/blackarch/tree/patch-8), the proposed PR is updated automatically as well? Ok, I will consider it for the next tasks. Thank you very much.

noraj: Yeah, the same apply if you work on the CLI.

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md doc/*.md
+
+  rm -rf LICENSE *.md doc/*.md .gitignore
+
+  cp -a * "$pkgdir/usr/share/$pkgname/"
+
+  cat > "$pkgdir/usr/bin/$pkgname" << EOF
+#!/bin/sh
+source /usr/share/$pkgname/venv/bin/activate
```
noraj, line 36: despite the fact that all dependencies were added, traces of the previous venv setup are remaining

noraj (changes requested):

packages/osintgram/PKGBUILD
D3vil0p3r: The package seems to work correctly without that venv line. The only issue should be inside the bin file that we generate at the end of package(). Osintgram looks for username in `config/credentials.ini`. Our package will set this file in `/usr/share/osintgram/config`.

The problem is: if I run `osintgram` command in a directory without `config/credentials.ini` it won't find it and will ask always for username. So, it works only if we run it in a directory with `config/credentials.ini` directory/file. Need to find a way to force osintgram to refer always to `/usr/share/osintgram/config` directory.

noraj: Yes unfortunately often we have to use `cd /usr/share/$pkgname` in wrappers because tools call relative files in a wrong manner.

D3vil0p3r: Y. I implemented it.

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+package() {
+  cd $pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
```
noraj, line 27: this is unneeded

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md doc/*.md
+
+  rm -rf LICENSE *.md doc/*.md .gitignore
```
noraj, line 30:
```suggestion
  rm -rf LICENSE *.md doc/ .gitignore Makefile Dockerfile docker-compose.yml docker_reqs.txt
```

noraj (changes requested):

packages/osintgram/PKGBUILD
D3vil0p3r: Fixed

D3vil0p3r:
Hey @noraj at the state of the art of this PR, it is needed to keep the label "do-not-merge" and apply some changes or it is all good?

noraj:
> Hey @noraj at the state of the art of this PR, it is needed to keep the label "do-not-merge" and apply some changes or it is all good?

The label was there because we were waiting for the dependencies to be merged. I'll check if everything is ready now.

noraj:
This paackage was not tested, it doesn't build:

image

packages/osintgram/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+
+pkgver() {
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+package() {
+  cd $pkgname
+
```
noraj, line 25:
```suggestion

  install -dm 755 "$pkgdir/usr/bin"
  install -dm 755 "$pkgdir/usr/share/$pkgname"

```

https://github.com/BlackArch/blackarch-pkgbuilds/blob/d57c25f19cd48b8a5af21d7948a509db8d9fdc92/PKGBUILD-generic-jointscript#L32-L33

noraj (changes requested):
Those are the missing lines making the build fail

noraj:
With the required changes, it's working properly

image

noraj:
You can use this cheat sheet https://gist.github.com/noraj/5045be32490f1161161e16fc5820066c and the tool ba-dev https://github.com/Edu4rdSHL/blackarch-devtools to test you package in a minimal reproducible environment.

D3vil0p3r:
> You can use this cheat sheet https://gist.github.com/noraj/5045be32490f1161161e16fc5820066c and the tool ba-dev https://github.com/Edu4rdSHL/blackarch-devtools to test you package in a minimal reproducible environment.

Thanks @noraj , it works. In general, when a PR is merged, when the tool will be compiled and available in BA repository?

noraj:
One has to add the package name into `list/to-release` like I did in https://github.com/BlackArch/blackarch/commit/3e28fd03a3f3472f1d78ac24cb072646ebb09b18 in order to track the packages that have to be released (new package) or re-released (package updated). Then noptrix from times to times (but quite often) manually check this file to trigger the build pipeline that will create the `.tar.zst` tarball archive and deploy it on the root repository so that mirrors will be available to sync with it.

D3vil0p3r:
> 

Ok thank you. What I don't see is the update release commit on `to-release` file about `evil-winrm` we fixed.

PR #3656, opened 2023-01-03
snscrape: add package

D3vil0p3r:
Scraper for social networking services (SNS). It scrapes things like user profiles, hashtags, or searches and returns the discovered items, e.g. the relevant posts.

Source: https://github.com/JustAnotherArchivist/snscrape


packages/snscrape/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+
+pkgname=snscrape
+pkgver=v0.4.3.20220106.r73.gcbdaee6
+pkgrel=1
+pkgdesc='A social networking service scraper in Python.'
+arch=('any')
+groups=('blackarch' 'blackarch-recon' 'blackarch-social')
+url='https://github.com/JustAnotherArchivist/snscrape'
+license=('GPL3')
+depends=('libxml2' 'libxslt' 'python' 'python-lxml')
```
noraj, line 12: Looks to me that you are missing the python deps https://github.com/JustAnotherArchivist/snscrape/blob/cbdaee6864aa8d3e9e4326d4fe224a517aabbc2c/setup.py#L27-L31

Also it don't looks like all othe one you puts are needed.

packages/snscrape/PKGBUILD
```diff
@@ -0,0 +1,33 @@
...
+makedepends=('git')
+source=("git+https://github.com/JustAnotherArchivist/$pkgname.git")
```
noraj, line 14: https://pypi.org/project/snscrape/

use this template instead https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-setuptools

noraj (changes requested):

packages/snscrape/PKGBUILD
D3vil0p3r: Just curiosity @noraj : why do you prefer to retrieve the file from pypi instead of GitHub project repo?

packages/snscrape/PKGBUILD
D3vil0p3r: Added needed dependencies: 'python-filelock' 'python-lxml' 'python-requests' 'python-beautifulsoup4'

I didn't add pytz due to apply only to < Python 3.9.

According to https://pypi.org/project/snscrape/#description , lxml needs libxslt and libxml2. I removed libxslt because already installed by python-lxml package, and removed also libxml2 because installed when libxslt is installed.

packages/snscrape/PKGBUILD
D3vil0p3r: PS: used the python-setuptools and implemented the necessary dependencies.

noraj: > Just curiosity @noraj : why do you prefer to retrieve the file from pypi instead of GitHub project repo?

Because we can keep integrity check, we can check for dependency changes at each release, and we made a script to auto-bump on new release for pypi releases cf. https://github.com/BlackArch/blackarch/blob/master/scripts/up-libraries

packages/snscrape/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+_pkgname=snscrape
+pkgver=0.4.3.20220106
+pkgrel=1
+pkgdesc='A social networking service scraper in Python.'
+arch=('any')
+groups=('blackarch' 'blackarch-recon' 'blackarch-social')
+url='https://github.com/JustAnotherArchivist/snscrape'
+license=('GPL3')
+depends=('python' 'python-beautifulsoup4' 'python-filelock' 'python-lxml' 'python-requests')
+makedepends=('git' 'python-setuptools')
```
noraj, line 14:
```suggestion
makedepends=('python-setuptools')
```

packages/snscrape/PKGBUILD
D3vil0p3r: Thanks

packages/snscrape/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+source=("https://files.pythonhosted.org/packages/source/${_pkgname::1}/$_pkgname/$_pkgname-$pkgver.tar.gz")
+sha512sums=('SKIP')
```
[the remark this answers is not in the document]
D3vil0p3r: Thank you for the suggestion. Applied SHA512 hash.

Issue #2759, opened 2020-05-06
Broken package monitoring


noraj:
[broken packages](https://github.com/BlackArch/blackarch/blob/master/lists/broken) reason

- [ ] aimage: `configure: error: OpenSSL library 'libssl-dev' not installed; cannot continue.`
- [ ] amoco: #2760
- [ ] autopwn: upstream https://github.com/nccgroup/autopwn/issues/14
- [ ] aws-extender-cli & python-boto https://github.com/BlackArch/blackarch/issues/4141

aancw:
[] birp: downgrade python2-py3270 to 0.3.4

```
 △ ~ sudo pacman -S birp
resolving dependencies...
looking for conflicting packages...

Packages (3) python2-py3270-0.3.5-4  x3270-1:3.6ga8-1  birp-65.b2e108a-1

 ▲ ~ python2                                                                                                                                                                                                                                                                 23s
Python 2.7.18 (default, Apr 23 2020, 22:32:06)
[GCC 9.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from py3270 import x3270App
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name x3270App

 ▲ ~ sudo pacman -Rns python2-py3270                                                                                                                                                                                                                                          9s
checking dependencies...
error: failed to prepare transaction (could not satisfy dependencies)
:: removing python2-py3270 breaks dependency 'python2-py3270' required by birp


 ▲ ~ python2                                                                                                                                                                                                                                                                 10s
Python 2.7.18 (default, Apr 23 2020, 22:32:06)
[GCC 9.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from py3270 import x3270App
>>>
```

noptrix:
> * [x]   birp: downgrade python2-py3270 to 0.3.4
> 
> ```
>  △ ~ sudo pacman -S birp
> resolving dependencies...
> looking for conflicting packages...
> 
> Packages (3) python2-py3270-0.3.5-4  x3270-1:3.6ga8-1  birp-65.b2e108a-1
> 
>  ▲ ~ python2                                                                                                                                                                                                                                                                 23s
> Python 2.7.18 (default, Apr 23 2020, 22:32:06)
> [GCC 9.3.0] on linux2
> Type "help", "copyright", "credits" or "license" for more information.
> >>> from py3270 import x3270App
[...]

PR #3662, opened 2023-01-04
Added angr-management

D3vil0p3r:
Adding amoco, angr-management because dependent on pyside2 (Python2-based)

### DIFF
--- a/lists/python2-tools
+++ b/lists/python2-tools
@@ -7,11 +7,13 @@ against
 albatar
 allthevhosts
 altdns
+amoco
 analyzemft
 analyzepesig
 androbugs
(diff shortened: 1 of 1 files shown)


noraj:
amoco is broken see https://github.com/BlackArch/blackarch/issues/2759 and https://github.com/BlackArch/blackarch/blob/master/lists/broken, so asking for a rebuild won't do

D3vil0p3r:
> amoco is broken see #2759 and https://github.com/BlackArch/blackarch/blob/master/lists/broken, so asking for a rebuild won't do

Ok. Removed.

Issue #3659, opened 2023-01-04
LinkScope


D3vil0p3r:
- **Name**: LinkScope
- **Source URL**: https://github.com/AccentuSoft/LinkScope_Client
- **Description**: LinkScope Client Online Investigation software allows you to perform online investigations by representing information as discrete pieces of data, called Entities.

I prepared two different types of PKGBUILD for it, one calling the `build.sh` script that build and install all the needed packages by pip, and another one where I deleted pip and venv commands. This last type of PKGBUILD I made was a little mess because of an important number of python module packages not stored in Arch repositories (along with their sub-deps), and PySide6 dependency gives error during building.

The first PKGBUILD type appears like the following:
```
# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
# See COPYING for license details.

pkgname=linkscope
_pkgname=LinkScope_Client
pkgver=v1.4.0.r22.g118b9c2
pkgrel=1
pkgdesc='Perform online investigations by representing information as discrete pieces of data, called Entities.'
arch=('any')
groups=('blackarch' 'blackarch-forensic' 'blackarch-recon')
url='https://accentusoft.com/downloads'
license=('GPL3')
depends=('python')
makedepends=('git' 'patchelf' 'python-pip' 'python-wheel')
source=("git+https://github.com/AccentuSoft/$_pkgname.git")
sha512sums=('SKIP')

pkgver() {
  cd $_pkgname

  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
}

build() {
  cd $_pkgname

  source build.sh
}

package() {
  cd $_pkgname

  install -dm 755 "$pkgdir/usr/bin"
  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md

  rm -rf LICENSE *.md .gitignore

  cp -a * "$pkgdir/usr/share/$pkgname/"

  cat > "$pkgdir/usr/bin/$pkgname" << EOF
#!/bin/sh
source /usr/share/$pkgname/venv/bin/activate
exec python /usr/share/$pkgname/LinkScope.py "\$@"
EOF

  chmod a+x "$pkgdir/usr/bin/$pkgname"
}
```
The compilation and the package compression take a good number of hours, the resulting tool should have a size of 3.5 GB due to a big number of recon modules used by the tool.

[...]

PR #3664, opened 2023-01-06
python-aiosmtplib: added package

D3vil0p3r:

### DIFF
--- a/packages/python-aiosmtplib/PKGBUILD
+++ b/packages/python-aiosmtplib/PKGBUILD
@@ -10,7 +10,7 @@ arch=('any')
 url='https://github.com/cole/aiosmtplib'
 license=('MIT')
 depends=('python')
-makedepends=('python-setuptools')
+makedepends=('python-build' 'python-pip')
 options=(!emptydirs)
 source=("https://files.pythonhosted.org/packages/source/${_pkgname::1}/$_pkgname/$_pkgname-$pkgver.tar.gz")
 sha512sums=('SKIP')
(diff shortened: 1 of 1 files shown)


packages/python-aiosmtplib/PKGBUILD
```diff
@@ -0,0 +1,30 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-aiosmtplib
+_pkgname=aiosmtplib
```
noraj, line 5:
```suggestion
_pkgname=${pkgname#python-}
```

packages/python-aiosmtplib/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+pkgver=2.0.0
+pkgrel=1
+pkgdesc='Asynchronous SMTP client for use with asyncio.'
+arch=('any')
+url='https://github.com/cole/aiosmtplib'
+license=('MIT')
+depends=('python')
+makedepends=('python-setuptools')
```
noraj, line 13: do not use deprecated setup-tools when possible, upstream has a https://github.com/cole/aiosmtplib/blob/main/pyproject.toml, use https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib-PEP517 instead

noraj (changes requested):

packages/python-aiosmtplib/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+
+  python setup.py build
+}
+
+package() {
+  cd "$_pkgname-$pkgver"
+
+  python setup.py install --root="$pkgdir" --prefix=/usr -O1 --skip-build
+  
+  rm -rf "$pkgdir/usr/lib/python3.10/site-packages/tests"
```
noraj, line 29:
```suggestion
  # workaround https://github.com/cole/aiosmtplib/issues/195
  rm -rf "$pkgdir/usr/lib/python3.10/site-packages/tests"
```

noraj (changes requested):

packages/python-aiosmtplib/PKGBUILD
D3vil0p3r: PKGBUILD-python-lib-PEP517 template applied.

PR #3665, opened 2023-01-06
python-docx2python: added package

D3vil0p3r:


packages/python-docx2python/PKGBUILD
```diff
@@ -0,0 +1,28 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-docx2python
+_pkgname=docx2python
```
noraj, line 5:
```suggestion
_pkgname=${pkgname#python-}
```

packages/python-docx2python/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+pkgver=2.0.5
+pkgrel=1
+pkgdesc='Extract docx headers, footers, text, footnotes, endnotes, properties, and images to a Python object.'
+arch=('any')
+url='https://pypi.org/project/docx2python'
```
noraj, line 10:
```suggestion
url='https://github.com/ShayHill/docx2python'
```

packages/python-docx2python/PKGBUILD
```diff
@@ -0,0 +1,28 @@
...
+license=('MIT')
+depends=('python' 'python-lxml')
+makedepends=('python-setuptools')
+options=(!emptydirs)
+source=("https://files.pythonhosted.org/packages/source/${_pkgname::1}/$_pkgname/$_pkgname-$pkgver.tar.gz")
+sha512sums=('SKIP')
```
noraj, line 16: dont skip, skip is only for git

noraj (changes requested):

packages/python-docx2python/PKGBUILD
D3vil0p3r: Applied SHA512 hash

PR #3672, opened 2023-01-06
python-Wappalyzer: added package

D3vil0p3r:

### DIFF
--- a/packages/python-Wappalyzer/PKGBUILD
+++ b/packages/python-Wappalyzer/PKGBUILD
@@ -1,7 +1,7 @@
 # This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
 # See COPYING for license details.
 
-pkgname=python-Wappalyzer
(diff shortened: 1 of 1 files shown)


packages/python-Wappalyzer/PKGBUILD
```diff
@@ -0,0 +1,28 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=python-Wappalyzer
```
noraj, line 4:
```suggestion
pkgname=python-python-wappalyzer
```

never put upper case in pkgname also the convention is `python-<pythonpkgname>` even if that makes  `python-python-something`

noraj (changes requested):

D3vil0p3r:
Probably this PR should be closed. Duplicated: https://github.com/BlackArch/blackarch/tree/master/packages/python-python-wappalyzer

PR #3679, opened 2023-01-11
cansina - remove fuzzdb dependency

D3vil0p3r:
fuzzdb is not a dependency, it is a wordlist set. On GitHub tool project, there is no requirement about fuzzdb. I would delete it because a user should be free to choose the wordlist set, and avoid to have "bloat" resources. Furthermore, it can create conflict issues if a user get the FuzzDB wordlist on its own by FuzzDB GitHub. In my opinion it is better to not use a wordlist repo as FuzzDB as "dependency", because it is not mandatory for the correct working of the tool.


noraj (changes requested):
LGTM

noraj (approved):

PR #3682, opened 2023-01-13
manticore - package updated

D3vil0p3r:


D3vil0p3r:
Waiting for https://github.com/trailofbits/manticore/issues/1426 to be fixed

PR #3685, opened 2023-01-13
WIP: angr-management: fix pkgbuild

noraj:
- pkgver++
- fix source
- closes #3662
- fix license
- fix build() and package() as setup.py was removed in profit to pyproject.toml and setup.cfg
- fix dependencies
- python-pyside6-qtads can't be built for now because shiboken6-generator is missing see https://bugs.archlinux.org/task/77117 and https://github.com/mborgerson/pyside6_qtads/blob/main/pyproject.toml#L9

### DIFF
--- a/packages/angr-management/PKGBUILD
+++ b/packages/angr-management/PKGBUILD
@@ -2,23 +2,21 @@
 # See COPYING for license details.
 
 pkgname=angr-management
-pkgver=9.1.11752
+pkgver=9.2.33
 pkgrel=1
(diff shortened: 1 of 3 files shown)


D3vil0p3r:
Hey. I just reworked on the package and I opened a new PR. I used the venv approach due to the number of dependencies to manage. What do you think? https://github.com/BlackArch/blackarch/pull/4661

noraj:
venv is a good idea in this case

D3vil0p3r:
At this point this PR can be closed

PR #3704, opened 2023-01-28
Remove python-flask-cors and python-flask-socketio

D3vil0p3r:
These two packages are duplicated of python2-flask-cors and python2-flask-socketio. The related PKGBUILD have been removed but the related packages were not from the repository. Please, remove them because they are in conflict with the correct python2-related version ones.

The affected packages should be `rspet` and `trape` (this last one use as dependency `python-flask-cors` and `python-flask-socketio` that should be Python3 and not cloned Python2 deps.

### DIFF
--- a/lists/to-remove
+++ b/lists/to-remove
@@ -1,2 +1,4 @@
+python-flask-cors - (PKGBUILD deleted but the pkg file still exists. This package is a duplicated of python2-flask-cors)
+python-flask-socketio - (PKGBUILD deleted but the pkg file still exists. This package is a duplicated of python2-flask-socketio)
 smartphone-pentest-framework - (deprecated - unmaintained > 6 years)
 zeus-scanner - (multiple upstream issues, unmaintained > 2 years)


noraj:
[python-flask-cors](https://github.com/BlackArch/blackarch/blob/master/packages/python-flask-cors/PKGBUILD) is py3 so I don't see how it is redundant with a py2 package [python2-flask-cors](https://github.com/BlackArch/blackarch/blob/master/packages/python2-flask-cors/PKGBUILD). The only thing I see is that pkgname below should be `python-flask-cors` as both have the same name.

https://github.com/BlackArch/blackarch/blob/3cce7ddd00deb55696314d040447754675698222/packages/python2-flask-cors/PKGBUILD#L4

Actually we should merge both packages in the same dual PKGBUILD like for [python-python-nmap](https://github.com/BlackArch/blackarch/blob/master/packages/python-python-nmap/PKGBUILD).

For [python-flask-socketio](https://archlinux.org/packages/community/any/python-flask-socketio/) it's an AL package, it's not actually in BA.

For [python2-flask-socketio](https://github.com/BlackArch/blackarch/blob/master/packages/python2-flask-socketio/PKGBUILD), same the issue here is the pkgname:

https://github.com/BlackArch/blackarch/blob/3cce7ddd00deb55696314d040447754675698222/packages/python2-flask-socketio/PKGBUILD#L4

Issue #3708, opened 2023-02-07
Remove/rename a tool called snitch


tocic:
Please remove/rename a tool called `snitch` since [the upstream URL](http://ntsecurity.nu/toolbox/snitch) doesn't exist anymore. It redirects to a tool collection which doesn't have a similar tool AFAIK. According to the Wayback Machine, it was designed for Windows 95 / 98 / ME / NT 4.0 and has now been removed.

The actual problem is that its name clashes with a name of [a different package from AUR](https://aur.archlinux.org/packages/snitch) which causes some inter-repository crawling issues.

tocic:
@noptrix any updates on this?

tocic:
@D3vil0p3r Hello. Could you please look at this?

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/3984

D3vil0p3r:
@noraj look the linked PR above. Should we close this issue?

tocic:
Are there any "discussion thread about this or better via matrix"?

noraj:
> @noraj look the linked PR above. Should we close this issue?

Yes.

PR #3709, opened 2023-02-09
snitch: remove package

tocic:
Closes https://github.com/BlackArch/blackarch/issues/3708.


noraj:
This is not how removing a tool works, I would rather be adding it to this list: https://github.com/BlackArch/blackarch/blob/master/lists/to-remove

tocic:
@noraj OK, and how long will it take to remove the package then? (There is a package in this list that was requested to be removed on Jun 9, 2021.)

noraj:
I'm not sure as I don't manage this part of the infrastructure.

PR #3710, opened 2023-02-11
sherlock: version bump

D3vil0p3r:
The current Sherlock in BlackArch uses `username_unclaimed` attribute so, when it is used, you will receive an error. This issue is solved on the latest version of Sherlock.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+sherlock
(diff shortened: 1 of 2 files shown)


noraj (approved):

noptrix:
@D3vil0p3r thx

PR #3712, opened 2023-02-11
crackmapexec-poetry: add package

noraj:
as crackmapexec is a dependency nightmare and it's nearly impossible to maintain a package for it without system dependency issue, I suggest an alternative package using a virtual environment with poetry


0xF61 (commented):
I'm using pipx version of myself for the user. The poetry solution seems working system-wide without dependency issues. Can you add the master?

Issue #2842, opened 2020-09-08
Read this, before posting new tool request!


ghost:
### Reasons for closing or having closed a tool request:
1. It is a simple wrapper for another tool, for example, a wrapper for Nmap that merely calls the program with pre-configured keys.
2. The tool is very outdated or no longer supported by its developers, which often creates dependency problems for other projects.
3. The tool has a broken build, relies on complex variables, or is highly dependent on a specific GNU/Linux distribution (not Arch Linux).
4. The tool is based on Python 2, and Python 2 support has ended, leading to issues with its dependencies.
5. The tool duplicates the functionality of existing tools and can be entirely replaced by a more popular and supported product.
6. Utilities related to anonymization or various scripts for launching free VPNs are not necessary since we provide [torctl](https://github.com/BlackArch/torctl)
7. Any tools related to illegal or unethical activities, such as brute-forcing Facebook accounts or conducting Denial of Service (DOS) attacks, will not be added.
### Please read this carefully before submitting a new tool request!

saltasatelites:
Blakarch has lots of tools with python2 dependencies. Sure the devs have think about that, but I suposse is a huge work.

noraj:
I put it in the issue template instead: https://github.com/BlackArch/blackarch/commit/4cb8df0800e37ef36384f917439b48ee960ec0dd

PR #3717, opened 2023-02-19
Add package : MHHDoS 

PardesiCat:
Requesting you to add [MHDDoS](https://github.com/MatrixTM/MHDDoS) a open source ddos attack tool in blackarch
i already wrote the PKGBUILD    
 + I used official pkgbuild template
+ its python3 based tool
+ template used [PKGBUILD-python-standalone](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-standalone)

you can check my pkgbuild code here separately for easy inspect 

+ [My pkgbuild ](https://github.com/PardesiCat/blackarch-ddosattck-pkgbuild)
+ [My pkgbuild on my own git server](https://git.pardesicat.xyz/pardesicat/blackarch-ddosattck-pkgbuild)

### DIFF
--- a/packages/MHDDoS/MHDDoS.install
+++ b/packages/MHDDoS/MHDDoS.install
@@ -1,0 +1,13 @@
+post_install() {
+  set -e
+  cd /usr/share/MHDDoS
+  python -m venv venv
+  source venv/bin/activate &&
+    pip install --isolated --root="/usr/share/MHDDoS" --prefix='venv' \
+      -r requirements.txt
+    pip install -e .
+}
+
+post_upgrade() {
+  post_install "$@"
+}
--- a/packages/MHDDoS/PKGBUILD
+++ b/packages/MHDDoS/PKGBUILD
@@ -1,0 +1,43 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+
+pkgname=MHDDoS
+pkgver=2.4.1
+pkgrel=1
+pkgdesc='Best DDoS Attack Script Python3, (Cyber / DDos) Attack With 56 Methods.'
+arch=('any')
+groups=('blackarch' 'blackarch-dos')
+url='https://github.com/MatrixTM/MHDDoS'
+license=('MIT')
+depends=('python3')
+makedepends=('git' 'python-setuptools' 'python-pip')
+source=("git+https://github.com/MatrixTM/$pkgname.git")
+sha512sums=('SKIP')
+install="$pkgname.install"
+
+pkgver() {
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
+}
+
+package() {
+  cd $pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md
+
+  install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+
+  rm -rf LICENSE *.md .gitignore
+
+  cp -a * "$pkgdir/usr/share/$pkgname/"
+
+  cat > "$pkgdir/usr/bin/$pkgname" << EOF
+#!/bin/sh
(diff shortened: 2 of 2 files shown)


noraj:
Generically speaking we don't package DDoS tools, I think the only exception in our repository is Ddosify. Because those tools are very rarely used for ethical or legit purpose and more generally used only by script kiddies for illegal activities.

I'll let the others judge if we accept this tool or not and just technically review your PR in the meanwhile. If we refuse DDoS tools I'll update #2842 to prevent people PRing this kind of tools in the future.

packages/MHDDoS/MHDDoS.install
```diff
@@ -0,0 +1,13 @@
+post_install() {
```
noraj, line 1: there can't be uppercase in the file path

packages/MHDDoS/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+
+pkgname=MHDDoS
+pkgver=2.4.1
+pkgrel=1
+pkgdesc='Best DDoS Attack Script Python3, (Cyber / DDos) Attack With 56 Methods.'
```
noraj, line 6: description shouldn't be promotional eg. using _best_ but rather neutral

packages/MHDDoS/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+arch=('any')
+groups=('blackarch' 'blackarch-dos')
+url='https://github.com/MatrixTM/MHDDoS'
+license=('MIT')
+depends=('python3')
```
noraj, line 11: this dependency doesn't exist

```suggestion
depends=('python')
```

packages/MHDDoS/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+
+pkgname=MHDDoS
```
noraj, line 3: package name must be lower cased

packages/MHDDoS/MHDDoS.install
```diff
@@ -0,0 +1,13 @@
...
+  set -e
+  cd /usr/share/MHDDoS
+  python -m venv venv
+  source venv/bin/activate &&
+    pip install --isolated --root="/usr/share/MHDDoS" --prefix='venv' \
```
noraj, line 6: looking at https://github.com/MatrixTM/MHDDoS/blob/main/requirements.txt this package doesn't seem to require a virtual environment and a jointscript, all dependencies are in AL or BA

noraj (changes requested):

packages/MHDDoS/PKGBUILD
PardesiCat: i copied exactly fro the discription  becoz tool owner is not mine  but i will keep it mind

packages/MHDDoS/PKGBUILD
PardesiCat: aa sorry i  am new to pkgbuild   i saw python2 ended support so i put 3  .. i will take note

packages/MHDDoS/PKGBUILD
PardesiCat: oke got you.......

packages/MHDDoS/MHDDoS.install
PardesiCat: ayy got it  

but we need this git+https://github.com/MHProDev/PyRoxy.git
i think its not avalable in BA or AR

noraj: You're right, tihs one is not, in this case it need to be packaged with this template https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-lib-setuptools, it should be pretty straightforward.

PardesiCat:
> I'll let the others judge if we accept this tool or not and just technically review your PR in the meanwhile. If we refuse DDoS tools I'll update https://github.com/BlackArch/blackarch/issues/2842 to prevent people PRing this kind of tools in the future.

any update on this ?   or should i continue to the work as per review  changes and again do the PR?

noraj:
You can already do the changes previously discussed on a new commit to update this PR.

PR #3720, opened 2023-02-24
kerberoast: version bump

D3vil0p3r:
It solves the conflict issues due to the usage of `tests` folder in the root of Python3.10 folder. It has been solved on https://github.com/skelsec/kerberoast/pull/9


noraj (approved):
LGTM

PR #3733, opened 2023-03-17
feroxbuster: version bump

0xF61:


noptrix:
@0xF61 thanks for this, but no needed to open PRs for single vcs packages sicne we do upgrades (mostly) regularly every week. a new package upgrade run will be performed today. thanks again.

PR #3739, opened 2023-03-29
sagan: Update PKGBUILD

D3vil0p3r:
Replacing `mysql-clients` dependency by `mariadb-clients` because the former creates conflicts with other packages. `mariadb-clients` provides the same functionality of `mysql-clients` and is compatible with older version of MySQL.


noraj (approved):
LGTM

noptrix:

PR #3742, opened 2023-04-01
shosubgo: add package

D3vil0p3r:


noraj:
I'm not sure this tool is super useful since there are already tons of OSINT tools doing subdomain enumeration using Shodan API.

D3vil0p3r:
> I'm not sure this tool is super useful since there are already tons of OSINT tools doing subdomain enumeration using Shodan API.

It is used as dependency of ars0n framework, a bug bounty hunting framework that I would like to pack for BlackArch: https://github.com/R-s0n/ars0n-framework

noraj:
In that case it could be ok.

PR #3747, opened 2023-04-02
batman-adv: version bump

D3vil0p3r:


noptrix:
@D3vil0p3r thank you - btw, please don't delete the last empty line. there should be always a `\n` in each PKGBUILD.

PR #3753, opened 2023-04-02
python-pysctp: moved linux-headers dependency to optdepends

D3vil0p3r:
According to the project page, there is no request about the usage of `linux-headers` package. The install of this package could be dangerous for those systems using different kernel types. it could break the system.

The needed dependencies for pysctp are specified at https://github.com/P1sec/pysctp/blob/master/README.txt or https://github.com/P1sec/pysctp/blob/master/Makefile in `installdeps` rule.

Furthermore, python2 logic removed because deprecated and security risk.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-pysctp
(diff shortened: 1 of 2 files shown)


noptrix:
@D3vil0p3r thank you. please don't remove the `python2` package as other tools (for example: `sigploit`) are depending on that.

D3vil0p3r:
> @D3vil0p3r thank you. please don't remove the `python2` package as other tools (for example: `sigploit`) are depending on that.

Hey @noptrix . As requested, I recovered `python2` package lines of code on PKGBUILD.

noptrix:

PR #3755, opened 2023-04-02
scamper: version bump

D3vil0p3r:
scamper updated and removed `linux-headers` dependency because no needed.


noptrix:
@D3vil0p3r thank you

PR #3766, opened 2023-04-06
mrkaplan: add package

D3vil0p3r:
https://github.com/Idov31/MrKaplan


packages/mrkaplan/PKGBUILD
```diff
@@ -0,0 +1,21 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=mrkaplan
+pkgver=1.1.1
+pkgrel=1
+pkgdesc='Help red teamers to stay hidden by clearing evidence of execution.'
+arch=('any')
+groups=('blackarch' 'blackarch-windows' 'blackarch-exploitation' 'blackarch-recon')
```
noraj, line 9:
```suggestion
groups=('blackarch' 'blackarch-windows' 'blackarch-exploitation')
```

not related to recon

packages/mrkaplan/PKGBUILD
```diff
@@ -0,0 +1,21 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=mrkaplan
+pkgver=1.1.1
```
noraj, line 5: Maybe use git tags for automatic updates

noraj (changes requested):

packages/mrkaplan/PKGBUILD
D3vil0p3r: > Maybe use git tags for automatic updates

I used the tag version in order to remain in line with pkgbuild-windows BA PKGBUILD template for downloading directly the tool from "Release" section.

packages/mrkaplan/PKGBUILD
D3vil0p3r: > 

Since it is a "moonwalk fork" but for Windows, since Moonwalk reports in its README that one of the features is the [Reconnaissance](https://github.com/mufeedvh/moonwalk), is it correct or not to provide also `blackarch-recon` group?

noraj: We don't have to trust all what upstream says, plus the word is used in a different meaning

packages/mrkaplan/PKGBUILD
noraj: The templates are some bases but not strict boundaries, sometimes they can be mixed up. We'll see what noptrix choose.

packages/mrkaplan/PKGBUILD
D3vil0p3r: Removed `blackarch-recon`.

packages/mrkaplan/PKGBUILD
D3vil0p3r: > The templates are some bases but not strict boundaries, sometimes they can be mixed up. We'll see what noptrix choose.

@noraj I think your approach is better. Automating it can improve the efficiency of managing the package. I updated the PKGBUILD code, but I have only one doubt: 

since `pkgver` has a format like `Vx.y.z.commitinfo` and the `source` needs only `Vx.y.z`, I created a `_pkgver` that takes the `pkgver` and remove the string related to the commit, in order to have only `Vx.y.z`. The problem is that it must be done manually at each update/commit of the tool. In your opinion there is a better strategy for automating `_pkgver` value?

packages/mrkaplan/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+_pkgver=${pkgver//.r1.g6119344}
+pkgrel=1
+pkgdesc='Help red teamers to stay hidden by clearing evidence of execution.'
+arch=('any')
+groups=('blackarch' 'blackarch-windows' 'blackarch-exploitation')
+url='https://github.com/Idov31/MrKaplan'
+license=('BSD2')
+depends=()
+makedepends=()
+source=("$pkgname::git+https://github.com/Idov31/MrKaplan.git"
```
noraj, line 15:
```suggestion
source=("$pkgname::git+https://github.com/Idov31/$pkgname.git"
```

it's case-insensitive, don't worry

noraj (changes requested):

packages/mrkaplan/PKGBUILD
noraj: I'm sorry, I just realized that https://github.com/Idov31/MrKaplan/releases/download/V1.1.1/MrKaplan_Standalone.ps1 is different from https://github.com/Idov31/MrKaplan/blob/master/MrKaplan.ps1, it includes the external modules as base64 strings to be 100% standalone, and there is no automated way to generate it from the script in the git repository. So there is no advantage of cloning the git repository. I thought we would be able to copy MrKaplan.ps1 or generate the standalone version from it. In that case your earlier packaging was good.

D3vil0p3r: @noraj I was thinking to implement `pkgver()` function as:
```
pkgver() {
  cd $pkgname
  
  git describe --tags --abbrev=0
}
```
in order to get directly the tag string used for the source to be downloaded, but when I apply it, pkgver variable does not change automatically. Do you know the reason? It works only if I use `git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'`?

noraj: Since, in fact we don't use anything from the git repository a git package makes no sense, only a fixed version and fetching the github release script is fine as you did initially. My bad I made you change it, I thought the script in release was the same as in the repository.

D3vil0p3r: Applied on https://github.com/BlackArch/blackarch/pull/3787 because the current is already merged.

PR #3768, opened 2023-04-06
eyewitness: moving firefox dependency to optdepends

D3vil0p3r:

### DIFF
--- a/packages/eyewitness/PKGBUILD
+++ b/packages/eyewitness/PKGBUILD
@@ -15,7 +15,7 @@ depends=('python' 'python-pyqt4' 'python-netaddr' 'ghost-py' 'python-rdpy'
          'python-fuzzywuzzy' 'python-pillow' 'geckodriver'
          'python-easyprocess' 'python-pytesseract')
 makedepends=('git')
-optdepends=('firefox')
+optdepends=('firefox: Selenium module support')
 source=("$pkgname::git+https://github.com/FortyNorthSecurity/EyeWitness.git")
 sha512sums=('SKIP')
 
--- a/packages/eyewitness/PKGBUILD
+++ b/packages/eyewitness/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=eyewitness
 pkgver=1017.4d7736e
-pkgrel=1
+pkgrel=2
 groups=('blackarch' 'blackarch-webapp' 'blackarch-recon' 'blackarch-misc')
 pkgdesc='Designed to take screenshots of websites, provide some server header info, and identify default credentials if possible.'
 arch=('any')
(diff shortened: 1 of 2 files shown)


packages/eyewitness/PKGBUILD
```diff
@@ -12,9 +12,10 @@ license=('GPL3')
 depends=('python' 'python-pyqt4' 'python-netaddr' 'ghost-py' 'python-rdpy'
          'python-selenium' 'python-rsa' 'xorg-server-xvfb'
          'python-beautifulsoup4' 'python-pyvirtualdisplay'
-         'python-fuzzywuzzy' 'python-pillow' 'geckodriver' 'firefox'
+         'python-fuzzywuzzy' 'python-pillow' 'geckodriver'
          'python-easyprocess' 'python-pytesseract')
 makedepends=('git')
+optdepends=('firefox: Selenium module support')
```
noraj, line 18: Seems like only firefox is expected

https://github.com/FortyNorthSecurity/EyeWitness/blob/4d7736e18c019a3d3bd43d4f44f212fcae532a3b/Python/setup/setup.sh#L8-L9

https://github.com/FortyNorthSecurity/EyeWitness/blob/4d7736e18c019a3d3bd43d4f44f212fcae532a3b/Python/setup/setup.sh#L66

D3vil0p3r: Currently Yes, but checking the code it seems firefox executable is used mainly for Python Selenium module (I used `grep -rnw ./ -e "firefox"`) . Furthermore, the code tries to install firefox-esr that has a different executable file than firefox, so I think it is a small error of the developer.

noraj: But selenium drives the headless browser required by eyewitness to work right. Can you try to build and run it without firefox on a clean environment with ba-dev to be sure the move is safe?

D3vil0p3r: No firefox binary:
```
#                                  EyeWitness                                  #
################################################################################
#           FortyNorth Security - https://www.fortynorthsecurity.com           #
################################################################################

Starting Web Requests (1 Hosts)
Message: Expected browser binary location, but unable to find binary in default location, no 'moz:firefoxOptions.binary' capability provided, and no binary flag set on the command line

Finished in 0.2802412509918213 seconds
[*] No report files found to open, perhaps no hosts were successful
```
If I use another browser, for example `brave` and I rename its binary as `firefox`, the output is:
```
################################################################################
#                                  EyeWitness                                  #
################################################################################
#           FortyNorth Security - https://www.fortynorthsecurity.com           #
################################################################################

Starting Web Requests (1 Hosts)
Message: Process unexpectedly closed with status 0

Finished in 1.4661085605621338 seconds
[*] No report files found to open, perhaps no hosts were successful
```
If `firefox` is installed, it works correctly.

PR #3769, opened 2023-04-06
findsploit: moving firefox dependency to optdepends

D3vil0p3r:


noraj:
Have you tested the change on a minimal environment?

PR #3770, opened 2023-04-06
firefox-security-toolkit: moving firefox dependency to optdepends

D3vil0p3r:
In case a user using a browser different from firefox (i.e., firefox-esr or brave), if it installs `blackarch-misc` group package, it won't have two different browsers installed.

I would follow the same approach of `firefox-decrypt`: if the user is expecting to install a package related to firefox, it should already have it installed.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+firefox-security-toolkit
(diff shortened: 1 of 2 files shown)


noraj:
This installs extensions to firefox so FF is required.

PR #3771, opened 2023-04-06
sn1per: moving firefox dependency to optdepends

D3vil0p3r:


packages/sn1per/PKGBUILD
```diff
@@ -23,6 +23,7 @@ depends=('amap' 'arachni' 'bash' 'bind' 'brutex' 'cisco-torch' 'curl' 'jq'
          'testssl.sh' 'theharvester' 'unzip' 'w3af' 'wapiti' 'whatweb'
          'whois' 'wpscan' 'xprobe2' 'xsstracer' 'xdg-utils')
 makedepends=('git')
+optdepends=('firefox')
```
noraj, line 26: Can you add the short description: https://wiki.archlinux.org/title/PKGBUILD#optdepends

noraj (changes requested):

packages/sn1per/PKGBUILD
D3vil0p3r: Added.

noraj:
Have you tested it on a minimal environment?

PR #3795, opened 2023-04-13
semgrep: update and rework

noraj:
- Upgrade to 1.17.1
- Update dependencies
- `ubuntu-16.04.tgz` is no longer available since 1.15.x so have to fetch `semgrep-core` from wheel
- it should be possible to build it with ocaml like in https://github.com/Homebrew/homebrew-core/blob/5a4fd4fdb399dbd7d659ce738d81ba868e38a162/Formula/semgrep.rb#L179-L217 but would be more complex
- from setup.py to pyproject.toml
- remove unused old patch


noptrix:
@noraj thank you very much

PR #3796, opened 2023-04-14
python-zconfig: add package

D3vil0p3r:
It is a dependency of `pwncat-caleb` tool. Currently, `pwncat-caleb` cannot be installed because this dependency is missing from repository.


noptrix:
@D3vil0p3r thank you!

PR #3798, opened 2023-04-15
python-pip-shims: add package

D3vil0p3r:
python-pip-shims has been removed from Arch repository and moved to AUR and it is used as dependency for some BlackArch tool.


noptrix:
@D3vil0p3r thank you!

PR #3799, opened 2023-04-16
python-pysctp: fixed linux-headers in package()

D3vil0p3r:
On the previous PR https://github.com/BlackArch/blackarch/pull/3753 I forgot to remove linux-headers from the package() functions since they have been already specified in optdepends at the beginning.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-pysctp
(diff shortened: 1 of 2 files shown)


noptrix:
@D3vil0p3r thank you

PR #3816, opened 2023-04-22
sentrypeer: add package

D3vil0p3r:
Protect SIP Servers from bad actors.


D3vil0p3r:
@noptrix did you have the chance to check this?

PR #3821, opened 2023-04-23
recoverdm: add package

D3vil0p3r:
This tool is old but it uses simple libraries and it still works. It is used on new Kali Purple as purple team tool.

Feel free to add it or not.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+recoverdm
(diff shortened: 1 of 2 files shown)


noraj:
I'm not a fan of using a pre-built binary from a web archive. Is there a source code repository where we can build the tool from?

D3vil0p3r:
> I'm not a fan of using a pre-built binary from a web archive. Is there a source code repository where we can build the tool from?

Currently no, it is the only one.

PR #3828, opened 2023-05-05
[cpp2il] Update Cpp2IL to the latest commit at this date + Suggestion

Fijxu:
This tool is being updated more often since the `new-analysis` branch is now the default.

I would like to suggest a change in the current `pkgver()` method from the PKGBUILD, the date `2022.0.7`. This is because nobody is actually downloading the release `2022.0.7` of this software, they are downloading the latest commit (at the build date) from the default branch of Cpp2IL.

By the way. I can create an AUR package based on this PKGBUILD? I would like to have it there. (I should maintain the first 2 lines right?)

### DIFF
--- a/packages/cpp2il/PKGBUILD
+++ b/packages/cpp2il/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname=cpp2il
-pkgver=2022.0.7.r11.gdc9dcd9
+pkgver=2022.0.7.r14.g7a14115
 pkgrel=1
 epoch=1
 pkgdesc="A tool to reverse unity's IL2PP toolchain"


noraj:
You don't need to modify `pkgver()`, this is just for the naming of the version. It's not actually downloading the 2022 release, but the version is named like 2022 release + 14 commits since then.

noraj (approved):
LGTM

Fijxu:
> It's not actually downloading the 2022 release, but the version is named like 2022 release + 14 commits since then.

Yeah i know, i said that because some people can think this is an old release but is not, just that :p

noraj:
Merge, (I didn't test tho)

PR #3838, opened 2023-05-12
initial comment; rbac-lookup

dualfade:
Please review. Thank you !

### DIFF
--- a/packages/rbac-lookup/PKGBUILD
+++ b/packages/rbac-lookup/PKGBUILD
@@ -1,0 +1,44 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=rbac-lookup
+pkgver=237.0cfc20b
+pkgrel=1
+pkgdesc='RBAC Lookup is a CLI that allows you to easily find Kubernetes roles and cluster roles bound to any user.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-scanner' 'blackarch-recon')
+url='https://github.com/FairwindsOps/rbac-lookup'
+license=('Apache')
+depends=()
+makedepends=('git' 'go')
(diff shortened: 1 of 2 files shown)


packages/rbac-lookup/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+license=('Apache')
+depends=()
+makedepends=('git' 'go')
+source=("git+https://github.com/FairwindsOps/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo "$(git rev-list --count HEAD).$(git rev-parse --short HEAD)"
```
noraj, line 20: git tags available, please use the tags version

packages/rbac-lookup/PKGBUILD
```diff
@@ -0,0 +1,44 @@
...
+    -o $pkgname .
+}
+
+package() {
+  cd $pkgname
+
+  install -Dm 755 $pkgname "$pkgdir/usr/bin/$pkgname"
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
+
+  install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
```
noraj, line 42: Don't copy for generic license https://wiki.archlinux.org/title/PKGBUILD#license

It's actually mentioned in the template https://github.com/BlackArch/blackarch-pkgbuilds/blob/ed1aab7282e0d4143d3dfc62304376a3c946e705/PKGBUILD-go#L49-L51

noraj (changes requested):

packages/rbac-lookup/PKGBUILD
dualfade: Will fix.

packages/rbac-lookup/PKGBUILD
dualfade: My bad; will fix.

PR #3849, opened 2023-06-03
[Tool Request]: dirforce

guendouzaimed:
hi,
this is the fastest directory/file brute force tool.
dirforce is made with c++ and use thread to send multiple HTTP requests in the fastest way.
this is the too repo:
https://github.com/guendouzaimed/dirforce


noraj:
There are already tons of Web directory and file scanner tools.

image

Many are pretty fast and have more features. I don't see the interest of adding one more alternative with less features that is pretty new and unused.

Issue #3811, opened 2023-04-21
GooFuzz


m3n0sd0n4ld:
- **Name**: GooFuzz
- **Source URL**: https://github.com/m3n0sd0n4ld/GooFuzz
- **Description**: GooFuzz is a tool to perform fuzzing with an OSINT approach, managing to enumerate directories, files, subdomains or parameters without leaving evidence on the target's server and by means of advanced Google searches (Google Dorking).

D3vil0p3r:
Tool already submitted. @noraj can you close this?

m3n0sd0n4ld:
Sorry, I didn't know I had to close it.

Thank you again for including the tool :D

Take care!

PR #3855, opened 2023-06-12
add GooFuzz

marianomarciello:
Add GooFuzz as requested in #3811


noraj (approved):

noptrix (changes requested):
@marianomarciello Thank you for this. But a couple of things: I don't see any entries in `depends=()` but you use at least `curl`. We need to make sure that we add any external tools as dependencies inside `depends=()` you use in this shell script. Second thing: Please remove the name of the tool inside `pkgdesc`. You can say "A script written in Bash..." or sth similar.

marianomarciello:
Hi @noptrix thank your for your revision and for your time.

noptrix (approved):

m3n0sd0n4ld:
Thank you for add GooFuzz tool!

Issue #3859, opened 2023-06-19
pwntools is creating conflicting errors due to python-pwntools


D3vil0p3r:
BlackArch repo contains `pwntools` but Arch Linux as [python-pwntools](https://archlinux.org/packages/extra/any/python-pwntools/) package in Extra repository that is used as dependencies in several infosec packages and this creates conflicts with `pwntools`.

Since `pwntools` and `python-pwntools` come from the same source, could be suitable to delete `pwntools` from BlackArch repo?

noraj:
For now `sudo pacman -Syu --overwrite='/usr/lib/python3.11/site-packages/pwnlib/*' --overwrite='/usr/lib/python3.11/site-packages/pwntools-4.10.0-py3.11.egg-info/*' --overwrite='/usr/lib/python3.11/site-packages/tests/*
' --overwrite='/usr/lib/python3.11/site-packages/pwn/*'` is required to upgrade if one has a tool requiring pwntools. After the upgrade: `sudo pacman -Rns pwntools`.

noraj:
#3868 was handled, this issue can be closed?

D3vil0p3r:
> #3868 was handled, this issue can be closed?

Yes. Closed.

PR #3862, opened 2023-06-19
adenum: fixed pwntools dependency

D3vil0p3r:
Changed pwntools to python-pwntools: https://github.com/BlackArch/blackarch/issues/3859


D3vil0p3r:
Not clear why pkgcheck fails about `Python 2.7`.

noptrix:
@D3vil0p3r thank you

PR #3868, opened 2023-06-22
delete: pwntools

D3vil0p3r:
[python-pwntools](https://archlinux.org/packages/extra/any/python-pwntools/) added in Arch Linux Extra repository on May 2023. BA `pwntools` no need anymore. All tools dependent on `pwntools` have been switched to `python-pwntools`.

It solves https://github.com/BlackArch/blackarch/issues/3859

### DIFF
--- a/lists/to-remove
+++ b/lists/to-remove
@@ -1,2 +1,3 @@
+pwntools - (added in Arch Linux Extra repository)
 smartphone-pentest-framework - (deprecated - unmaintained > 6 years)
 zeus-scanner - (multiple upstream issues, unmaintained > 2 years)


ikstream (approved):
@noptrix I think it's a good idea to remove the package as it is upstream now and we don't have to care for conflicts

noraj (approved):
Relying on AL package will be better.

D3vil0p3r:
Thank you guys, we replaced `pwntools` depedency with the one of AL repository and PRs have been merged by @noptrix . If you see other BA packages using the `pwntools` dep, please let me know.

noptrix:
done

PR #3872, opened 2023-06-23
fixed: python-dsinternals

D3vil0p3r:
Temporary fix of tests conflicting error.

Ref: https://github.com/p0dalirius/pydsinternals/pull/11

Thank you for the PR on the source project @noraj

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-dsinternals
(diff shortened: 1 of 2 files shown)


noraj:
@D3vil0p3r Why would you add the RM since the upstream PR was merged? We just need to bump the package then.

D3vil0p3r:
> @D3vil0p3r Why would you add the RM since the upstream PR was merged? We just need to bump the package then.

Because the source of the package is not -git and I am afraid that the dsinternal developer will publish the new and fixed PyPi package too late.

PR #3876, opened 2023-06-26
remove braces

noraj:


noraj:
I guess it broken when there was a force push, I don't even remember what it was doing, my bad for not putting a description to the PR

PR #3814, opened 2023-04-22
undbx: add package

D3vil0p3r:
Extract e-mail messages from Outlook Express DBX files.

PR #3879, opened 2023-06-27
undbx: add package

noraj:
replace #3814


noptrix:
@noraj @D3vil0p3r next time: fix correct `arch` type: as long as nothing gets compiled -> `any`, as soon as something gets compiled -> `'x86_64' 'aarch64'`

PR #3883, opened 2023-06-28
remove useless make dependencies n°1

noraj:
https://wiki.archlinux.org/index.php/PKGBUILD#makedepends

The dependencies required should be the minimum and so eliminating useless dependencies.

> Note: The meta-package [base-devel](https://archlinux.org/packages/core/any/base-devel/) is assumed to be already installed when building with makepkg. Members of this group should not be included in `makedepends` array.

> Tip: The following can be used to check whether a particular package is either in the base-devel group or is pulled in by a member of the group: 
>
> `LC_ALL=C pacman -Si $(pactree -rl package) 2>/dev/null | grep -q "^Groups *:.*base-devel"`

Commands to search for useless deps:

```
grep -r "^makedepends=(" packages | grep -E "'(autoconf|automake|binutils|bison|fakeroot|file|findutils|flex|gawk|gcc|gettext|grep|groff|gzip|libtool|m4|make|pacman|patch|pkgconf|pkg-config|sed|sudo|texinfo|which)'"
```

Notes:

-  `glibc` is already pulled by some members of the `base-devel` group, maybe other sub-depencies may not be caught by the regexp
- `pkg-config` is an alias for `pkgconf` so other aliases may not be listed on the regexp
- with the example regexp, when makedepends is splitted on multiple lines, extra line won't be caught

duplicates should be removed too:

> The packages in the depends array are implicitly required to build the package, they should not be duplicated here.


packages/0d1n/PKGBUILD
```diff
@@ -3,15 +3,15 @@
...
-pkgrel=1
+pkgrel=2
 epoch=1
 groups=('blackarch' 'blackarch-webapp' 'blackarch-fuzzer' 'blackarch-scanner')
 pkgdesc='Web security tool to make fuzzing at HTTP inputs, made in C with libCurl.'
 arch=('x86_64' 'aarch64')
 url='https://github.com/CoolerVoid/0d1n'
 license=('GPL3')
 depends=('curl')
-makedepends=('git' 'make')
```
noraj, line 14: base-devel

packages/admsnmp/PKGBUILD
```diff
@@ -4,12 +4,11 @@
 pkgname=admsnmp
 _pkgname=ADMsnmp
 pkgver=0.1
-pkgrel=4
+pkgrel=5
 groups=('blackarch' 'blackarch-scanner')
 pkgdesc='ADM SNMP audit scanner.'
 license=('GPL')
 depends=('glibc')
-makedepends=('gcc')
```
noraj, line 12: base-devel

packages/afflib/PKGBUILD
```diff
@@ -4,15 +4,13 @@
...
 pkgver=3.7.19
-pkgrel=1
+pkgrel=2
 groups=('blackarch' 'blackarch-forensic')
 pkgdesc='An extensible open format for the storage of disk images and related forensic information.'
 arch=('any')
 url='https://github.com/sshock/AFFLIBv3'
 license=('custom:unknown')
 depends=('ncurses' 'zlib' 'openssl' 'curl' 'expat' 'readline' 'fuse2')
-makedepends=('make' 'gcc' 'ncurses' 'zlib' 'openssl' 'curl' 'expat' 'readline'
```
noraj, line 14: base-devel + duplicates with depends

packages/aiengine/PKGBUILD
```diff
@@ -3,15 +3,14 @@
...
-pkgrel=2
+pkgrel=3
 epoch=1
 groups=('blackarch' 'blackarch-networking' 'blackarch-defensive')
 pkgdesc='A packet inspection engine with capabilities of learning without any human intervention.'
 arch=('x86_64' 'aarch64')
 url='https://bitbucket.org/camp0/aiengine/downloads/'
 license=('GPL3')
 depends=('boost' 'boost-libs' 'python' 'libpcap')
-makedepends=('make')
```
noraj, line 14: base-devel

packages/avet/PKGBUILD
```diff
@@ -3,14 +3,14 @@
...
 pkgver=133.2f1d882
-pkgrel=2
+pkgrel=3
 pkgdesc='AntiVirus Evasion Tool'
 arch=('any')
 groups=('blackarch' 'blackarch-binary' 'blackarch-backdoor' 'blackarch-automation')
 url='https://github.com/govolution/avet'
 license=('GPL3')
 depends=('python' 'wine' 'wine-mono' 'wine_gecko')
-makedepends=('git' 'gcc')
```
noraj, line 13: base-devel

D3vil0p3r:
One question: can `base-devel` be used as **makedeps**? If so or not, why?

noraj:
> One question: can `base-devel` be used as **makedepends**? If so or not, why?

I think meta groups can be used as dependencies yes. But as explained in the ArchLinux wiki, it's useless to include it.

PR #3890, opened 2023-07-01
add: ciphey

D3vil0p3r:
It is an interesting project: https://github.com/Ciphey/Ciphey/

I inserted it along with the needed dependencies. The only issue is that it does not work with Python3.11 currently but with Python3.7 so I added it as dependency.

@noraj is there a way to copy these python installed dependencies from `/usr/lib/python3.11/site-packages/` to `/usr/lib/python3.7/site-packages/`?


D3vil0p3r:
I just realize on their Discord server that the project has ceased and migrated to another tool in Rust called Ares: https://github.com/bee-san/Ares. I will create a package for it.

PR #3892, opened 2023-07-01
sslyze: Fix sslyze and dependency

ikstream:
python-tls-parser:
-  hasn't been build for python 3.11 which made sslyze fail. 
- remove python2 instructions as no package in our repo is using it

sslyze:
- was missing a dependency (python-pyopenssl).

change both packages to PEP517


noptrix:
@ikstream thank you

PR #3893, opened 2023-07-02
python-peid: add package

astro-angelfish:
I found this simple tool accidentally when I was trying to reverse engine a program. So I decided to add this.

Seems to be duplicated with https://github.com/BlackArch/blackarch/tree/master/packages/packerid, but this tool works on Python3.
And `packerid` seems to be out of order on the malformed PE file and was removed from the source website. Should I open an issue about that?


packages/python-peid/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+  cat src/peid/VERSION.txt
+}
+
+build() {
+  cd $pkgname
+
+  pip wheel .
+}
+
+package() {
```
noraj, line 29: please use a template from https://github.com/BlackArch/blackarch-pkgbuilds, [PKGBUILD-python-PEP517](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-python-PEP517) as there is a [pyproject.toml](https://github.com/packing-box/peid/blob/main/pyproject.toml)

packages/python-peid/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+  cd $pkgname
+  pip install "./$_pyname-$pkgver-py3-none-$arch.whl" --root $pkgdir
+  install -Dm 644 LICENSE -t "${pkgdir}/usr/share/licenses/${pkgname}"
```
noraj, line 32: Only install license if required according to  https://wiki.archlinux.org/index.php/PKGBUILD#license

packages/python-peid/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+license=('GPL3')
+depends=('python' 'python-pefile' 'python-pip')
+makedepends=('python-setuptools')
+source=("$pkgname::git+https://github.com/packing-box/peid")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  cat src/peid/VERSION.txt
```
noraj, line 20: if a git package use one of https://github.com/BlackArch/blackarch-pkgbuilds/blob/8c072641adc8c758c4de8babb9285faee0501cd4/PKGBUILD-generic-git#L17-L25

packages/python-peid/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+pkgname=python-peid
+_pyname=peid
+pkgver=1.2.9
+pkgrel=1
+pkgdesc='Python implementation of the Packed Executable iDentifier (PEiD)'
+arch=('any')
+grouos=('blackarch' 'blackarch-binary' 'blackarch-reversing')
+license=('GPL3')
+depends=('python' 'python-pefile' 'python-pip')
+makedepends=('python-setuptools')
```
noraj, line 13: no need of it if you don't use it

noraj (changes requested):

packages/python-peid/PKGBUILD
astro-angelfish: I have no idea if `python-setuptools` really needs to be removed. But in the project.toml it seems to be a build requirement in https://github.com/packing-box/peid/blob/main/pyproject.toml#L2

packages/python-peid/PKGBUILD
astro-angelfish: The version of peid uses `src/peid/VERSION.txt` according to https://github.com/packing-box/peid/blob/main/src/peid/__info__.py#L17-L18
I'm really confused which kind of version should be used to meet the version rule of BlackArch.

noraj: when pypi package source we use a fixed version (eg. semver) but when the source is git, it's mandatory to use one of the two command I linked, for auto-updates and more.

packages/python-peid/PKGBUILD
noraj: You can test with and without it to see if it's really needed. https://gist.github.com/noraj/5045be32490f1161161e16fc5820066c#testing

astro-angelfish: I've realized it does not need `python-setuptools` but instead `python-build` is required now. Thanks for making me aware about that.

packages/python-peid/PKGBUILD
astro-angelfish: Now I've migrated this package to pypi since it is available in pypi, and followed the version from pypi.
I was discovered this project in GitHub and mistakenly made this package as a git package.

PR #3900, opened 2023-07-06
crackmapexec upgrade

### LINKED ISSUE #3877: Upgrade crackmapexec
CME moved of repository and v6.0.0 is out

- source: https://github.com/Porchetta-Industries/CrackMapExec ➡️ https://github.com/mpgn/CrackMapExec
- version 5.4.0 ➡️ 6.0.0

Ref.

- https://wiki.porchetta.industries/news/a-new-home
- https://twitter.com/mpgn_x64/status/1673445546294493189
- https://github.com/Porchetta-Industries/CrackMapExec/blob/master/README.md

noraj:
fix #3877

---

https://github.com/BlackArch/blackarch/commit/91d57106c879d30276db240c91ab34379b03fbfe python-python-libnmap is no longer compatible with python2, but 2 old packages abandoned for 6-8 years are still relying on the python2 version

- [ ] lhf
- [ ] atear

If we want to keep a python 2 version maybe we should use a separate package for the py2 and p3 and apply a patch on the py2.

---

upgraded cme, `python-dploot` is missing but it was not an issue as far as I tested

https://github.com/BlackArch/blackarch/commit/28bcb54ce9f0425e257dbc3e84eb54b5d335f2ec

---

actually cme 6 is not building because the required sqlalchemy (https://github.com/mpgn/CrackMapExec/blob/beb6d9f73567780a173f4edd9d8f16b996d69a23/pyproject.toml#L51) is requiring v2.0 but the one on AL is still at v1.4 and flagged out of date since 2023-05-23.


noptrix:
@noraj can you please provide the `python2-python-libnmap` in a separate packege with the old-version `0.7.3` then?

noraj:
Done ✅

PR #3908, opened 2023-07-07
facebrok: replaced mysql by mariadb

D3vil0p3r:
MariaDB is the official MySQL-related package. `mysql` package creates conflicts with other tools.


packages/facebrok/PKGBUILD
```diff
@@ -3,13 +3,13 @@
 
 pkgname=facebrok
 pkgver=33.0f6fe8d
-pkgrel=1
+pkgrel=2
 pkgdesc='Social Engineering Tool Oriented to facebook.'
 groups=('blackarch' 'blackarch-social')
-arch=('any')
+arch=('x86_64' 'aarch64')
```
noraj, line 9:
```suggestion
arch=('any')
```

this is not compiled

noraj (changes requested):

PR #3928, opened 2023-08-02
python-yenc: add package

D3vil0p3r:
yEnc is a binary-to-text encoding scheme for transferring binary files in messages on Usenet or via e-mail. The Python module can be also used on several Capture the Flags to solve InfoSec challenges by yEnc decoding.


noraj:
no update since 2014 and it's working with python 3.11 ?

D3vil0p3r:
> no update since 2014 and it's working with python 3.11 ?

Yes, it works. You can test it by solving this simple challenge: https://cybertalents.com/challenges/network/hidden-in-plain

Writeup for faster test: https://abdelrahman671.github.io/posts/Hidden-In-Plain/

Just note that for the writeup working script, you need the latest python-pyshark (I bumped the latest version on https://github.com/BlackArch/blackarch/pull/3929)

ikstream:
Have you checked the source code for obvious backdoors or similar? As it has zero states, a quick check would be good 

They still react to PRs, so I think it's fine to add it if the code is ok and it works.

D3vil0p3r:
> Have you checked the source code for obvious backdoors or similar? As it has zero states, a quick check would be good
> 
> They still react to PRs, so I think it's fine to add it if the code is ok and it works.

Yes I checked. There are no evidences of backdoors or similar. Did you experience a case where a proposed tool in BlackArch hid a backdoor?

D3vil0p3r:
@ikstream @noraj I'm reviewing all my open PRs. Did you analyze this one?

noraj:
I'm not competent to review C code.

PR #3940, opened 2023-08-19
Dnsrecon fix

PardesiCat:
dnsrecon not working throwing errors 

pcattestdnsrecon

but fixed  ( you can check the fixed  pkg + build log here )

https://git.pardesicat.xyz/pardesicat/dnsrecon-fix

problem  missing depends in PKGBUILD 

depends name on arch repo     **python-pytest**   **flake8**


noraj:
@PardesiCat Thanks for the PR. However, I don't think this is the right way to fix it.

pytest and flake8 are dev dependencies (for test and linting) so they are not required at runtime.

This dependency error is happening because they are set as required in the package. setup.cfg load the whole requirements.txt

https://github.com/darkoperator/dnsrecon/blob/38b705febfe37490cc9bbee10640d41553635fe6/setup.cfg#L2

requirements.txt mixes runtime and dev dependencies:

https://github.com/darkoperator/dnsrecon/blob/master/requirements.txt

So the solution is better packaging upstream. Eg. a separate requirements.txt and requirements-dev.txt (cf [link](https://improveandrepeat.com/2021/01/python-friday-55-separate-development-from-runtime-dependencies/)). There more options like `extras_require`, `tests_require` for stup.cfg. A better way would be to drop the obsolete setup.py anyway and migrate to pyproject.toml that offers more flexibility cf. https://python-poetry.org/docs/master/managing-dependencies/

noraj:
@PardesiCat Could you PR the changes upstream please?

PardesiCat:
well here also i  tested  in  for now     adding in   makedepends=('git' 'python-setuptools' 'pytest' 'flake8' )

also doing the job  so for now should we in BA  USe   in makedepends  ?

noraj:
Long term: No, because it's not a makedepends either. As a temporary workaround, I suggest doing all the following:

1. Add `python-pytest' 'flake8` in `makedepends`
2. Create an upstream issue or PR
3. Add a comment above the makedepends line of the PKGBUILD with the link to the issue or PR
4. (We'll be able to track this and remove the dependencies once upstream is fixed)

D3vil0p3r:
@noraj @PardesiCat https://github.com/BlackArch/blackarch/pull/3941

D3vil0p3r:
@noraj the source files seem to be fixed upstream. Should we remove the build deps from https://github.com/BlackArch/blackarch/pull/3941 ?

noraj:
@D3vil0p3r We could close #3941 as it was fixed in https://github.com/darkoperator/dnsrecon/commit/d0507ef2499d9cc3c3354e1c0632cfc5898a9bf3. However we need to bump pkgrel to fetch the fix.

D3vil0p3r:
> @D3vil0p3r We could close #3941 as it was fixed in https://github.com/darkoperator/dnsrecon/commit/d0507ef2499d9cc3c3354e1c0632cfc5898a9bf3. However we need to bump pkgrel to fetch the fix.

At this point should I keep the PR opened and bump only pkgrel? Or should I close it and @noptrix will bump pkgrel directly?

noraj:
Done ✅

PR #3947, opened 2023-09-05
nuclei: version bump

noraj:


noraj:
I was not sure if all tools where included in the monitor list or if it's 100% automatic.

PR #3958, opened 2023-09-26
add python-cerbeyra-api

D3vil0p3r:
This library provides a high level interface for interacting with Cerbeyra’s APIs and retrieving cyber-threat’s information from them. Through such API’s, it is possible to get clues regarding vulnerability assessments, cyber feeds, governance, threat intelligence and much more.

Just not sure about the `python-setuptools` dependency specified in its `pyproject.toml` file. This dependency could be still needed to build the package even if it is not based on setup.py?

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+python-cerbeyra-api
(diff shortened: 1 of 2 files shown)


noptrix:

PR #3968, opened 2023-10-06
ad-miner: release

noraj:


noptrix:
@noraj thanks for this. `pkgdesc` is missing, though :)

noraj:
> @noraj thanks for this. `pkgdesc` is missing, though :)

What the point of me putting a TODO if I don't check for remaining TODO before submitting 😂

PR #3969, opened 2023-10-09
modifycerttemplate: add package

dr-Styki:
[modifyCertTemplate](https://github.com/fortalice/modifyCertTemplate) is designed to aid an operator in modifying ADCS certificate templates so that a created vulnerable state can be leveraged for privilege escalation (and then reset the template to its previous state afterwards).

### DIFF
--- a/packages/modifycerttemplate/PKGBUILD
+++ b/packages/modifycerttemplate/PKGBUILD
@@ -1,0 +1,41 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=modifycerttemplate
+_pkgname=modifyCertTemplate
+pkgver=7.4c35708
+pkgrel=1
+pkgdesc='Aid operators in modifying ADCS certificate templates so that a created vulnerable state can be leveraged for privilege escalation.'
+arch=('any')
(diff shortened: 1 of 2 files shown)


packages/modifycerttemplate/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+
+package() {
+  cd $_pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+  install -dm 755 "$pkgdir/usr/share/$pkgname"
+
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
+
+  rm README.md .gitignore
```
noraj, line 31: We can remove `[requirements.txt](https://github.com/fortalice/modifyCertTemplate/blob/master/requirements.txt)` too.

noraj (changes requested):
One small remark else LGTM.

packages/modifycerttemplate/PKGBUILD
dr-Styki: Done

noraj (approved):
LGTM

noptrix:
@dr-Styki thank you

PR #3970, opened 2023-10-10
add exe2hex

D3vil0p3r:
Convert a Windows PE executable file to a batch file and vice versa. Ref:  https://www.kali.org/tools/exe2hexbat/
https://github.com/g0tmi1k/exe2hex/

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+exe2hexbat
--- a/packages/exe2hexbat/PKGBUILD
+++ b/packages/exe2hexbat/PKGBUILD
@@ -1,0 +1,42 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=exe2hexbat
+_pkgname=exe2hex
+pkgver=1.5.1
+pkgrel=1
+pkgdesc='Inline file transfer using in-built Windows tools (DEBUG.exe or PowerShell).'
+arch=('any')
+groups=('blackarch' 'blackarch-disassembler')
+url='https://github.com/g0tmi1k/exe2hex'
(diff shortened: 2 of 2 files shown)


noptrix:
@D3vil0p3r thanks for this. two things: why do we rename the package name? It shall stay like the original `exe2hex` unless file/package conflict happenes. is the `pkgdesc` accurate? i haven't checked the github page of `exe2hex`.

D3vil0p3r:
Thank you for the comment @noptrix 
I choose that pattern because it is the same used in Kali: https://www.kali.org/tools/exe2hexbat/

But if you agree, we can use exe2hex has pkg name.

noptrix:
> Thank you for the comment @noptrix I choose that pattern because it is the same used in Kali: https://www.kali.org/tools/exe2hexbat/
> 
> But if you agree, we can use exe2hex has pkg name.

oh, i see. well, we don't care about kali tbh... let's stick with its original name, since i don't see any packages which might conflict with that. :) also, please always add an empty line at the end of a `PKGBUILD` file, e.g. always a `\n`.

D3vil0p3r:
> > Thank you for the comment @noptrix I choose that pattern because it is the same used in Kali: https://www.kali.org/tools/exe2hexbat/
> > But if you agree, we can use exe2hex has pkg name.
> 
> oh, i see. well, we don't care about kali tbh... let's stick with its original name, since i don't see any packages which might conflict with that. :) also, please always add an empty line at the end of a `PKGBUILD` file, e.g. always a `\n`.

I agree. I changed the pkgname by the original one and added final `\n`.

noraj (approved):
LGTM

noptrix:
@D3vil0p3r thanks!

PR #3974, opened 2023-10-13
fixed unix-privesc-check

D3vil0p3r:
Fixed sha512 and binary name that was wrongly `unix-privsc-check`.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+unix-privesc-check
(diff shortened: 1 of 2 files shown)


noraj (commented):
I don't see a change for the binary name

D3vil0p3r:
> I don't see a change for the binary name

Hey. Try to install the current package. You will see that the binary name is the wrong one. I don't know why, indeed the old PKGBUILD (except for SHA512) seems right.

noraj:
kk so outside the hash it needs a rebuild

noptrix:
@D3vil0p3r thank you - please don't remove the last empty line inside a PKGBUILD.

PR #3985, opened 2023-10-23
hekatomb: add package

noraj:


noptrix:
@noraj please always add a `\n` at the end of a `PKGBUILD` file

noraj:
Yup, usually do, I must have edit this outside my usual editor because .editorconfig covers this

PR #3988, opened 2023-10-25
python-pyparsing dependency needed to provide QuotedString

SoothingDuck:
I tried to install wfuzz from a fresh blackarch on Arch Linux and got a "QuotedString" NameError exception.

A `sudo pacman -S python-pyparsing` solved the issue which is why I propose to include it in the deps of the package

### DIFF
--- a/packages/wfuzz/PKGBUILD
+++ b/packages/wfuzz/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=wfuzz
 pkgver=1155.1b695ee
-pkgrel=5
+pkgrel=6
 groups=('blackarch' 'blackarch-fuzzer' 'blackarch-webapp')
 pkgdesc='Utility to bruteforce web applications to find their not linked resources.'
 url='https://github.com/xmendez/wfuzz'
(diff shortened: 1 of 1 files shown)


packages/wfuzz/PKGBUILD
```diff
@@ -10,7 +10,7 @@ url='https://github.com/xmendez/wfuzz'
 arch=('any')
 license=('GPL2')
 depends=('python' 'python-lxml' 'python-pycurl' 'python-chardet' 'python-six'
-         'python-future' 'python-wxpython')
+         'python-future' 'python-wxpython' 'python-pyparsing')
```
noraj, line 13: It seems this dependency is listed and expected https://github.com/xmendez/wfuzz/blob/1b695ee9a87d66a7d7bf6cae70d60a33fae51541/setup.py#L34

noraj (changes requested):
LGTM, you just need to increment `pkgrel` please

SoothingDuck:
pkgrel incremented

noraj (approved):

noptrix:
@SoothingDuck thanks for this

PR #3990, opened 2023-10-27
#fix tor-router version bump

dualfade:
please review. Thanks.


noptrix:
please add `pkgver()` function

Issue #3996, opened 2023-10-28
hekatomb and yaaf conflicting


D3vil0p3r:
**hekatomb** and **yaaf** BA packages are in conflict:
```
/usr/lib/python3.11/site-packages/src/__init__.py exists in both 'hekatomb' and 'yaaf'
```
How can we fix it?

PS: yaaf seems to not be maintained anymore.

noraj:
It's not a real conflict, just bad upstream packaging. The solution is the same as for tests cf. https://github.com/BlackArch/blackarch/issues/3829#issuecomment-1536605202

I guess:

1. Temporary workaround: rm `/usr/lib/python3.11/site-packages/src/__init__.py` in the PKGBUILD
2. Long term fix: PR to exclude this file (example: [issue](https://github.com/cr0hn/enteletaor/issues/9), [PR](https://github.com/cr0hn/enteletaor/pull/10/files)

D3vil0p3r:
Hekatomb ok but yaaf seems to not be maintained anymore

noraj:
Everything in https://github.com/Processus-Thief/HEKATOMB/tree/main/src

![image](https://github.com/BlackArch/blackarch/assets/16578570/2f2c8797-1a39-4fa9-b251-0b17b0cdc25a)

is deployed under `usr/lib/python3.11/site-packages/` because the python package is poorly designed

noraj:
I'm working on a fix for yaaf

noraj:
- [x] fix yaaf https://github.com/BlackArch/blackarch/pull/3997
- [ ] fix hekatomb TODO

D3vil0p3r:
> * [x]  fix yaaf [yaaf: fix deployment #3997](https://github.com/BlackArch/blackarch/pull/3997)
> * [ ]  fix hekatomb TODO

Thank you. Just one question: if yaaf is not maintained anymore, why it is still kept on BlackArch repository?

noraj:
I think noptrix have explained it tons of time. There was also the idea of having an archive repository, etc. ask him on matrix if you want more details. But I think the idea is abandoned tool can be still relevant, and sometimes no activity is not an issue, a tool don't necessarily need more features over time and if it designed well there may be no need fix dependencies every week. So just because it is abandoned or archived doesn't mean it becomes useless, unmaintainable, irrelevant, etc.

D3vil0p3r:
[...]

PR #3997, opened 2023-10-28
yaaf: fix deployment

noraj:
- github username was renamed so it's redirecting to a new namespace
- fix bad upstream packaging resulting in conflicts cf. #3996


noptrix:
@noraj thanks for this

Issue #3791, opened 2023-04-11
[Tool Request] Anubis


wykwit:
- **Name**: Anubis
- **Source URL**: https://github.com/jonluca/Anubis
- **Description**: Anubis is a subdomain enumeration and information gathering tool. Anubis collates data from a variety of sources, including HackerTarget, DNSDumpster, x509 certs, VirusTotal, Google, Pkey, Sublist3r, Shodan, Spyse, and NetCraft.

PR #4000, opened 2023-11-01
Add anubis

wykwit:
See: #3791

### DIFF
--- a/packages/anubis/PKGBUILD
+++ b/packages/anubis/PKGBUILD
@@ -1,0 +1,48 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=anubis
+_pkgname=anubis-netsec
+pkgver=1.1.3.r0.g9ea89fa
+pkgrel=1
+pkgdesc='Subdomain enumeration and information gathering tool.'
+arch=('any')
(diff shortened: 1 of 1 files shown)


packages/anubis/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+depends=('python' 'python-python-nmap' 'python-shodan' 'python-docopt'
+         'python-requests' 'python-censys' 'python-dnspython')
+makedepends=('git' 'python-build' 'python-pip')
+source=("git+https://github.com/jonluca/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
```
noraj, line 22: There are git tags https://github.com/jonluca/Anubis/tags so you should use https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git#L24 instead

noraj (changes requested):

packages/anubis/PKGBUILD
wykwit: Isn't this exactly L24 of that template?

noraj (approved):
LGTM

noptrix:
@wykwit thanks for this

PR #4007, opened 2023-11-10
Fixed Tor-Browser package (again)

MarkusTieger:
Fixes that the arch variable doesn't get replaced in "tor-browser.in".

and btw simple question:
Do you even test this package? It's the second time, something broke.
(it hasn't found the tar.xz file, because the $PACKAGE_ARCH" was not replaced and because of that, it tryed to extract from a filepath, that does not exist)

### DIFF
--- a/packages/tor-browser/PKGBUILD
+++ b/packages/tor-browser/PKGBUILD
@@ -17,7 +17,6 @@ pkgrel=1
 pkgdesc='Tor Browser Bundle: anonymous browsing using Firefox and Tor.'
 groups=('blackarch' 'blackarch-defensive')
 url='https://www.torproject.org/projects/torbrowser.html'
-arch=('x86_64')
 license=('GPL')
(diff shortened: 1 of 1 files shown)


noptrix:
@MarkusTieger Thanks for this. Your feedback about testing is noted, and we'll improve our processes. Appreciate your contribution. Btw, `pkgrel++` next time you make changes if there is no version bump.

noptrix:
oh, i see why we didn't use it:
```
babuild: NOTE: package is not compatible with x86_64
copying built packages to '/home/noptrix/tmp/blackarch/tor-browser'...
cp: cannot stat '*.pkg.tar.zst': No such file or directory
ERROR: unable to read package file: '/home/noptrix/tmp/blackarch/tor-browser/*'.
ssh-agent already started.
waiting for lock...
cleaning repo: blackarch...
ARCH: x86_64
ARCH: aarch64
```
so we need to find another solution here

PR #4012, opened 2023-11-21
Update PKGBUILD (adfind)

sahakkhotsanyan:
after upgrading adfind from old php version to a new golang we also need to update PKGBUILD for properly install

### DIFF
--- a/packages/adfind/PKGBUILD
+++ b/packages/adfind/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname=adfind
-pkgver=v1.0.0
+pkgver=v1.0.2.r1.gee408bd
 pkgrel=1
 pkgdesc='Simple admin panel finder for php,js,cgi,asp and aspx admin panels.'
 groups=('blackarch' 'blackarch-webapp' 'blackarch-recon')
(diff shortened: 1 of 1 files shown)


packages/adfind/PKGBUILD
```diff
@@ -2,28 +2,31 @@
...
-makedepends=('git')
+makedepends=('git' 'go')
 source=("git+https://github.com/sahakkhotsanyan/$pkgname.git")
 sha512sums=('SKIP')
 
 pkgver() {
   cd $pkgname
 
-  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+  echo $(git describe --abbrev=0 --tags)
```
noraj, line 19: please use this instead:

https://github.com/BlackArch/blackarch-pkgbuilds/blob/581198db183f275d3d1f5b0135c94fb60df1b96a/PKGBUILD-generic-git#L24

packages/adfind/PKGBUILD
```diff
@@ -2,28 +2,31 @@
...
 }
 
 package() {
   cd $pkgname
-
+  
+  git checkout tags/$pkgver
```
noraj, line 25: Don't checkout.

packages/adfind/PKGBUILD
```diff
@@ -2,28 +2,31 @@
...
+  go build -o $pkgname ./cmd/adfind/main.go
```
noraj, line 26: Don't build in `package()` but in `build()`.

Use https://github.com/BlackArch/blackarch-pkgbuilds/blob/581198db183f275d3d1f5b0135c94fb60df1b96a/PKGBUILD-go#L27C1-L42

noraj (changes requested):
Thanks for the upgrade notification. However, some changes are required.

noraj (approved):
LGTM

PR #4013, opened 2023-11-24
Remove Ultimate Facebook Scraper

anunna:
The source code was removed. The developer confirmed he has close sourced the tool with commercial licensing.


SofianeHamlaoui (approved):

anunna:
@noptrix I get that we don't remove tools that are unmaintained, but the repo for this tool only provides a README.md because it was commercialized. The PKGBUILD is useless. The developer confirmed the same.
image

PR #4028, opened 2023-12-08
Checking for up-to-date pkg versions and update if necessary.

dante-tech:
The `update` function is a custom script that checks for updates for the installed packages from the `packages` and `aur` directories. It compares the local and remote versions of each package and prompts the user to install the updates if available.

To use the `update` function, simply run it in a terminal without any arguments. It will print a message indicating the progress of the check and the list of packages that have updates available. For example:

```
$ update
Checking for updates...
The following packages have updates available:
lib1
lib2
lib3
Do you want to install them? [Y/n]

```
The user can then enter `Y` or `y` to confirm the installation, `N` or `n` to cancel the installation, or any other input to abort the operation. The function will then call the `do_sync` function for each package that needs to be updated.

I hope you will like this, I wish everybody a great day!

### DIFF
--- a/scripts/blackman
+++ b/scripts/blackman
@@ -75,6 +75,45 @@ is_installed() {
 	pacman -Qi "$1" &> /dev/null
 }
 
+update() {
+  # Check for updates for the installed packages
+  printf "Checking for updates...\n"
+  local updates=()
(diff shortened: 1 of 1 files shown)


noptrix:
thank you, but obsolete.

PR #4052, opened 2024-01-07
add ntlm_theft

### LINKED ISSUE #4051: ntlm_theft
- **Name**: ntlm_theft
- **Source URL**: https://github.com/Greenwolf/ntlm_theft
- **Description**: generating multiple types of NTLMv2 hash theft files

noraj:
closes #4051

### DIFF
--- a/packages/ntlm_theft/PKGBUILD
+++ b/packages/ntlm_theft/PKGBUILD
@@ -1,0 +1,42 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=ntlm_theft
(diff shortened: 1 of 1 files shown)


noptrix:
@noraj we try to avoid using `_` and rather use `-` e.g. `tool-name`.  i will change that and i will also use the `misc` group rather than `networking`

noptrix:
i just saw, we had ntlm-theft(ntlm_theft) already in the repository.

noraj:
> @noraj we try to avoid using `_` and rather use `-` e.g. `tool-name`. i will change that and i will also use the `misc` group rather than `networking`

I thought of that but saw `create_ap`. But ok noted for next time.

PR #4062, opened 2024-01-23
cook: better pkgver

noraj:


noptrix:
@noraj you have to add `epoch=1`  because old ver > new ver (because of 'v')

PR #4105, opened 2024-03-22
dradis: quick build fix

noraj:
- [x] fix the build
  - [x] fix file no longer existing
  - [x] fix gemfile to allow not exactly the same ruby version
- [ ] fix runtime

Still have an issue with [sassc](https://github.com/sass/sassc-ruby) gem that is a binding for libsass

image


noptrix:
@noraj can we give it another try please? :)

noraj:
It doesn't run for other reasons now. Some files have changed. It's anoyingly complex to package, I mean it's a big web application.

Also, there is a Docker existing, I suggest using this instead https://github.com/dradis/dradis-ce/blob/develop/Dockerfile, in favor of https://github.com/BlackArch/blackarch/discussions/4004

PR #4114, opened 2024-03-30
paru: version bump

Ystri0n:


noptrix:
thanks - done via previous commit

PR #4120, opened 2024-04-13
mongodb: version bump

D3vil0p3r:


D3vil0p3r:
@noptrix @noraj how are you building mongodb PKGBUILD as resources in terms of RAM amount and how long it takes on your side?

packages/mongodb/PKGBUILD
```diff
@@ -16,7 +16,7 @@
 # Contributor: Fredy García <frealgagu at gmail dot com>
 
 pkgname=mongodb
-pkgver=7.0.5
+pkgver=7.0.8
```
noraj, line 19: sha512sums were not bumped.

noraj (changes requested):

PR #4125, opened 2024-04-26
Fixed aardwolf with version on aur

rad10:


rad10:
So @noptrix any reason why this merge request was denied? The package is failing, preventing any dependenants from working and it cant easily be replaced by aur because its a version ahead. If there is a change that would be preferred, I could comply and conform so that this package starts working for everyone again.

noraj:
I guess, it's because, rather than providing the required changes for the BA package, you simply pasted the AUR package.

PR #4127, opened 2024-04-26
Fixed crackmapexec with build from aur

rad10:


noraj (commented):
You can't just simply paste an AUR package here:

- For copyright reasons
- BA doesn't work the same
- BA has so syntax standards that are not always respected in random AUR packages

PR #4131, opened 2024-05-04
osert: fix pandoc dependency

D3vil0p3r:
Fixed dependency and target directory of pandoc. It solved the issue that prevented to generate PDF from input file.

@noraj


noraj (commented):
Thanks, I was not aware those dependencies were renamed and changed.

PR #4167, opened 2024-06-28
Add CDK

sawntoe:
Hi y'all! CDK is a zero dependency container penetration toolkit that I would really like to see added to BlackArch.

### DIFF
--- a/packages/cdk/PKGBUILD
+++ b/packages/cdk/PKGBUILD
@@ -1,0 +1,52 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=cdk
+pkgver=v1.5.3
+_pkgver=v1.5.3
+pkgrel=1
+pkgdesc='Zero Dependency Container Penetration Toolkit'
+arch=('any')
+groups=('blackarch' 'blackarch-scanner')
+url='https://github.com/cdk-team/CDK'
+depends=('bash')
+makedepends=()
+optdepends=()
+options=(!strip)
+source=(
+  'https://github.com/cdk-team/CDK/releases/download/v1.5.3/cdk_darwin_amd64'
+  'https://github.com/cdk-team/CDK/releases/download/v1.5.3/cdk_linux_386'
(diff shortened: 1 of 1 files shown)


packages/cdk/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+pkgdesc='Zero Dependency Container Penetration Toolkit'
+arch=('any')
+groups=('blackarch' 'blackarch-scanner')
+url='https://github.com/cdk-team/CDK'
+depends=('bash')
+makedepends=()
+optdepends=()
+options=(!strip)
+source=(
+  'https://github.com/cdk-team/CDK/releases/download/v1.5.3/cdk_darwin_amd64'
```
noraj, line 17: We don't distribute pre-built binaries, we build from source.
Use this PKGBUILD template: https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-go

noraj (changes requested):
Thanks for your PR, but a change will be required.

packages/cdk/PKGBUILD
sawntoe: Hi! Under any other situation, I would definitely agree with you. However, I took https://github.com/BlackArch/blackarch/blob/ab15c1a1607ff7a6f245f55be1033eceef86d52b/packages/peass/PKGBUILD as a template, and it does also seem to download pre-built binaries. In cases such as these where these binaries are not meant to be run on the host machine but distributed, it does seem to make more sense to use pre-distributed and trusted binaries, and not have to download cross-compile toolchains and packers for every architecture out there.

noraj: As you can read in the command for peass, it's because it was impossible to do so: sh2bin is not available for linpeas build (so it's a temporary measure intil we package [sh2bin](https://github.com/carlospolop/sh2bin)), and that winpeas build requires visual studio which is not available on linux so we have no choice. In your case it's different as it's a go binary, cross-compilation is way easier, there is no issue with gox dependency and upx is available as an official package.

PR #4171, opened 2024-07-04
Update mirror.lst

Hostico:
Hosting a blackarch mirror in Romania


noptrix:
@Hostico thank you for this. mind updating also on `blackarch-site.git` (`generators/downloads.html.sh`)?

noptrix:
@Hostico thank you very much - sorry totally missed it :)

PR #4199, opened 2024-09-01
sherlock: major packing update

### LINKED ISSUE #4191: Sherlock pkgbuild should be updated to reflect new versioning scheme
Not really a bug, not really a tool request...so here we are at discussion.

Sherlock [[upstream]](https://github.com/sherlock-project/sherlock) [[pkgbuild]](https://github.com/BlackArch/blackarch/blob/55abb0751c32e11ad84468f42fa0423c84d36543/packages/sherlock/PKGBUILD) was added quite a while ago, at which point the only true source of version info was found in the code itself. Not very conducive to consistent packaging. Due to this, Sherlock is packaged here on a rolling basis, similar to other distros (at the time).

Recently, Sherlock has adopted properly tagged releases somewhat styled after semver (not entirely, but similar). Most other repositories have adjusted [[repology]](https://repology.org/project/sherlock-social-osint/versions). Not only does this make for easier mental parsing of version information, it allows for much more consistent user experiences across platforms.

If able and compliant with the Black Arch standard, Sherlock's pkgbuild should be converted to use the now-available releases rather than the HEAD of master.

noraj:
- move to PEP517 packaging
- remove useless dependencies
- add missing dependency

also fix #4191

### DIFF
--- a/packages/sherlock/PKGBUILD
+++ b/packages/sherlock/PKGBUILD
@@ -2,46 +2,41 @@
 # See COPYING for license details.
 
 pkgname=sherlock
-pkgver=2467.f7075e1
+_pypi_name=sherlock-project
+_pypi_archive=sherlock_project
+pkgver=0.15.0
 pkgrel=1
 pkgdesc='Find usernames across social networks.'
 groups=('blackarch' 'blackarch-social' 'blackarch-recon')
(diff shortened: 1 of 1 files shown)


noptrix:
@noraj thank you

PR #4205, opened 2024-09-10
evil-winrm: Set environment variable instead of changing directory

Vz0n:
I added an extra environment variable to avoid changing directory when executing, because this causes issues with stuff like the download and upload function of the program.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+evil-winrm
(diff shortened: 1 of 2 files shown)


noptrix:
@Vz0n thank you!

noraj:
Nice idea, thanks.

Issue #4207, opened 2024-09-16
Tool Request - Villain


anunna:
- **Name**: Villain
- **Source URL**: https://github.com/t3l3machus/Villain
- **Description**: A C2 framework that manages and enhances reverse shells, sharing functionality across connected servers.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4212

PR #4212, opened 2024-09-24
villain: add package

D3vil0p3r:
Solve https://github.com/BlackArch/blackarch/issues/4207

### DIFF
--- a/packages/villain/PKGBUILD
+++ b/packages/villain/PKGBUILD
@@ -42,3 +42,4 @@ EOF
 
   chmod a+x "$pkgdir/usr/bin/$pkgname"
 }
+
--- a/packages/villain/PKGBUILD
+++ b/packages/villain/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname=villain
-pkgver=324.58a8c02
+pkgver=v2.2.0.r7.g58a8c02
 pkgrel=1
 pkgdesc='C2 framework that can handle multiple TCP socket & HoaxShell-based reverse shells, enhance their functionality with additional features and share them among connected sibling servers.'
 arch=('x86_64' 'aarch64')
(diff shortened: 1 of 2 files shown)


packages/villain/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+license=('CCBY-NC-ND-4.0')
+depends=('python' 'python-netifaces' 'python-pycryptodome' 'python-pyperclip' 'python-requests' 'python-gnureadline')
+makedepends=('git')
+source=("$pkgname::git+https://github.com/t3l3machus/Villain.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
```
noraj, line 20: Could you rather use that: https://github.com/BlackArch/blackarch-pkgbuilds/blob/a6f1e0daf6683ff186ac4fbd34fcd024a082c2f4/PKGBUILD-generic-git#L17-L23?
tags are availables

packages/villain/PKGBUILD
```diff
@@ -0,0 +1,42 @@
...
+}
+
+package() {
+  cd $pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
```
noraj, line 27: this file is not necessary and also you do a `cp -a` afterward

noraj (changes requested):

packages/villain/PKGBUILD
D3vil0p3r: Fixed

packages/villain/PKGBUILD
D3vil0p3r: Fixed

noraj (commented):
LGTM by eyes checking, one of use should test it in ba-dev before

PR #4214, opened 2024-09-26
fix zdns build, bluebox-ng sha

djustice:
it has a makefile


noptrix:
ok, i will add the `make` part. i will also update bluebox-ng sha. thank you. the rest on your PR: we use `$pkgname` and not the name directly. why do ou still have the plcscan/pmcma updates on this PR?

PR #4218, opened 2024-10-01
fix xxeserv build

djustice:
go.mod already existed, so build() errored out.


noptrix:
@djustice thanks a lot!

PR #4221, opened 2024-10-02
cubicsdr: add package

D3vil0p3r:
It needs at least one SOAPy package to be installed to work. Evaluate if move the related packages from optional to mandatory.

At start, the executable produces not-blocking messages like:
```
/usr/src/debug/wxwidgets/wxWidgets/src/unix/glegl.cpp(386): assert ""Assert failure"" failed in InitVisual(): Failed to get an EGLConfig for the requested attributes.
```
Reported on [AUR](https://aur.archlinux.org/packages/cubicsdr#comment-993142) and [upstream](https://github.com/cjcliffe/CubicSDR/issues/1037).

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+cubicsdr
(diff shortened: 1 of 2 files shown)


noptrix:
@D3vil0p3r thank you!

PR #4222, opened 2024-10-02
upgpkg: sickle

djustice:
build adjusted


noptrix:
@djustice thank you very much.

PR #4225, opened 2024-10-03
qbdi: add package

D3vil0p3r:
The upstream repository provides also [PyQBDI](https://pypi.org/project/PyQBDI/) but probably it could be created separately if needed.


noptrix:
@D3vil0p3r thank you!

Issue #3063, opened 2021-04-01
JD-GUI package refactoring


noraj:
Type: Discussion

## Discussion

I was wondering if [jd-gui PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/jd-gui/PKGBUILD) could be improved:

- migrating from gtk2 to gtk3?
- use 64 bits version instead of lib32?
- How to test with ba-dev since a X11 display is needed?
- are all the deps really needed?

aancw:
Is GTK required by JD-GUI? There is no GTK package in AUR AL https://aur.archlinux.org/packages/?O=0&SeB=nd&K=jd-gui&outdated=&SB=n&SO=a&PP=50&do_Search=Go , as far as I know it's using Swing.

Testing JD-GUI run in GTK3 and GTK2

```
#Gtk3 forced:
env SWT_GTK3=1 jd-gui
 
#Gtk2 forced:
env SWT_GTK3=0 jd-gui
```

All is running well 

![image](https://user-images.githubusercontent.com/6284204/114171796-fc12d400-995e-11eb-9c2e-85602ff95764.png)

> How to test with ba-dev since a X11 display is needed?

Can we redirect the display to Xserver? Is there any experience about testing GUI app? @noptrix @Edu4rdSHL

Edu4rdSHL:
@noraj you can simply set the `DISPLAY` variable and that should work. In my case I use a nspawn container (that has access to X server) and inside it the ba-dev environment. I used `ba-dev -p jd-gui-1.6.6-1-x86_64.pkg.tar.zst -e "DISPLAY=:0.0 jd-gui"` and it works. Anyways, that package is missing `libxtst` as dependency, without it you'll get: `libXtst.so.6:  cannot open shared object file: No such file or directory`

noraj:
So `gtk2` dependency is not needed?

Edu4rdSHL:
I didn't say that and I didn't even tested, I just said that with the actual PKGBUILD even the `libxtst` dep is missing.

aancw:
> So `gtk2` dependency is not needed?

Reference to AUR there is no gtk dependencies

noptrix:
@noraj any news on this? if not let's close it please

noraj:
idk, I don't want to modify what I don't understand

Tazmain:
Java swing shouldn't be dependant on GTK at all. Which environments do we want this to run in?

[...]

PR #4231, opened 2024-10-04
jd-gui: remove useless dependencies

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3063


noraj:
Is it still a draft?

D3vil0p3r:
Eh... This tool behaves in a weird manner with GTK envs. My dark GTK environment opens it as:
image

If I remove gtk-related packages, it opens as the default view. If I reinstall the same GTK packages I uninstalled, it remains as default view.

So, it actually depends on GTK, but not gtk2, because the system I'm testing on uses gtk3.

noptrix:
@D3vil0p3r thanks

Issue #3404, opened 2022-05-05
Instashell: source not found


qkqsb:
The source URL provided for instashell is unavailable. (https://github.com/thelinuxchoice/instashell)
The original author has seemingly disappeared without any new location for the code. I've linked what seems to be the most popular repo of this code that is still available below.
https://github.com/maxrooted/instashell

The package also cannot be downloaded from any mirror as far as I can tell. I tested the code in the repository I linked, and near as I can tell it works.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4235

PR #4235, opened 2024-10-05
instashell: replaced source

D3vil0p3r:
The replaced source has a number of commits lower than the non-existing one. It means that, before to publish this package, remove the old one.

Close https://github.com/BlackArch/blackarch/issues/3404


noraj:
> The replaced source has a number of commits lower than the non-existing one. It means that, before to publish this package, remove the old one.

You should increase epoch then. See https://wiki.archlinux.org/title/PKGBUILD#epoch

Issue #3631, opened 2022-12-16
pyinstxtractor


vdun:
- **Name**: pyinstxtractor
- **Source URL**: https://github.com/extremecoders-re/pyinstxtractor
- **Description**: PyInstaller Extractor

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4242

PR #4242, opened 2024-10-07
pyinstxtractor: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3631


noraj (approved):
LGTM, builds, runs, seems packaged correctly

Issue #3660, opened 2023-01-04
QT3 - Missing source files qt-copy-kde-patches.tar.bz2 and qt-patches.tar.bz2


D3vil0p3r:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

[qt3](https://github.com/BlackArch/blackarch/tree/master/packages/qt3) folder does not have two source files specified in its [PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/qt3/PKGBUILD). In particular:
* qt-copy-kde-patches.tar.bz2
* qt-patches.tar.bz2

For this reason, if you run `makepkg -si`, you will get the following error:
```
==> ERROR: qt-copy-kde-patches.tar.bz2 was not found in the build directory and is not a URL.
```
while, `PKGBUILD` has these files as source:
```
<SNIP>
source=("https://download.qt.io/archive/qt/3/qt-x11-free-$pkgver.tar.gz"
        'qt3-png15.patch' 'qt-copy-kde-patches.tar.bz2' 'qt-patches.tar.bz2'
        'eastern_asian_languagues.diff' 'qt-odbc.patch' 'gcc-4.6.patch'
        'qt-x11-free-3.3.5-makelibshared.patch')
<SNIP>
```

noraj:
Then the question is, is there still a tool that requires qt3?

D3vil0p3r:
> Then the question is, is there still a tool that requires qt3?

Currently, the BA package requiring qt3 are [pyqt3](https://github.com/BlackArch/blackarch/blob/master/packages/pyqt3) and [qscintilla-qt3](https://github.com/BlackArch/blackarch/blob/master/packages/qscintilla-qt3)

BA tools depending on pyqt3 are:
* [faradaysec](https://github.com/BlackArch/blackarch/tree/master/packages/faradaysec)
* [gerix-wifi-cracker](https://github.com/BlackArch/blackarch/blob/master/packages/gerix-wifi-cracker)
* [inguma](https://github.com/BlackArch/blackarch/tree/master/packages/inguma)

noraj:
> Currently, the BA package requiring qt3 are [pyqt3](https://github.com/BlackArch/blackarch/blob/master/packages/pyqt3) and [qscintilla-qt3](https://github.com/BlackArch/blackarch/blob/master/packages/qscintilla-qt3)

[...]

PR #4244, opened 2024-10-07
qt3: fix build issue

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3660


D3vil0p3r:
@noptrix be aware that [.gitignore](https://github.com/BlackArch/blackarch/blob/master/.gitignore) has `*.bz2` so it could impact those package sources with this kind of archives.

noraj:
uneeded, qt3 should be dropped

PR #4250, opened 2024-10-07
vpnpivot: to remove

D3vil0p3r:


noraj:
I already added it on the broken list https://github.com/BlackArch/blackarch/commit/bb470becd31caab9f632b1337129f7c0d291f427. Please don't PR on the to-remove list. noptrix could explain why. I understood it was reserved for urgencies or something like that.

Issue #3803, opened 2023-04-21
gsmevil2


utenus:
- **Name**: gsmevil2
- **Source URL**: https://github.com/ninjhacks/gsmevil2
- **Description**:
Useful when testing GSM networks.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4263

PR #4263, opened 2024-10-08
gsmevil2: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3803

### DIFF
--- a/packages/gsmevil2/PKGBUILD
+++ b/packages/gsmevil2/PKGBUILD
@@ -1,0 +1,46 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=gsmevil2
+pkgver=12.eb96b4f
+pkgrel=1
+pkgdesc='Python web-based tool which use for capturing imsi numbers and sms.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-radio' 'blackarch-sniffer')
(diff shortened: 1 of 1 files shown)


packages/gsmevil2/PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+pkgname=gsmevil2
+pkgver=12.eb96b4f
+pkgrel=1
+pkgdesc='Python web-based tool which use for capturing imsi numbers and sms.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-radio' 'blackarch-sniffer')
+url='https://github.com/ninjhacks/gsmevil2'
+license=('custom:unknown')
+depends=('gr-gsm' 'kalibrate-rtl' 'python-flask' 'python-flask-socketio' 'python-pyshark')
```
noraj, line 12: some dependencies are missing

image

please test installing and runing the package in minimal environment

either with ba-dev or https://wiki.archlinux.org/title/DeveloperWiki:Building_in_a_clean_chroot

noraj (changes requested):

packages/gsmevil2/PKGBUILD
D3vil0p3r: Sorry if I didn't use ba-dev in my last PRs. Now I retrieved it and fixed PKGBUILD according ba-dev output. Let me know if it's ok. Thanks

PR #4277, opened 2024-10-09
dnspy: fix target directory

D3vil0p3r:
Preventing copy of an intermediate directory in the target path because useless.


noptrix:
@D3vil0p3r thank you!

Issue #4087, opened 2024-02-22
levye -> crowbar


noraj:
levye has been renamed crowbar, the old repo https://github.com/galkan/levye no longer exist so need to migrate to https://github.com/galkan/crowbar

fields to update:

- url
- source
- check if depends or `package()` need updates
- change license to `license=('MIT')`

saltasatelites:
The 'new' repository has not had any updates in the last 4 years and is just another brute force tool of which there are already hundreds.

ikstream:
I think we should add/change it nevertheless to allow an (easier) transition from Kali to BA and keep these video guides made for Kali also work on BA

PR #4280, opened 2024-10-09
levye: broken

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4087

### DIFF
--- a/lists/broken
+++ b/lists/broken
@@ -50,6 +50,7 @@ inquisitor # discontinued, better delete
 katana-framework # discontinued, use python2, broken as relying on python2 dependencies that are not available in the repo
 keimpx
 lanmap2 # discontinued, better delete
+levye # source no longer exist, better delete
 lte-cell-scanner # discontinued, better delete
 malcom # #2857, #3572, discontinued, better delete
 manticore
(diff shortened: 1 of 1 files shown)


noraj:
I'm not sure about this one, ikstream said we should package [crowbar](https://github.com/galkan/crowbar) for compatibility with kali

D3vil0p3r:
crowbar has already its [PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/crowbar/PKGBUILD) in BA. Btw I submitted https://github.com/BlackArch/blackarch/pull/4327 to adapt it to the setup.py python PKGBUILD BA template. Feel free to accept or refuse it.

lists/broken
```diff
@@ -50,6 +50,7 @@ inquisitor # discontinued, better delete
 katana-framework # discontinued, use python2, broken as relying on python2 dependencies that are not available in the repo
 keimpx
 lanmap2 # discontinued, better delete
+levye # source no longer exist, better delete
```
noraj, line 53:
```suggestion
levye # source no longer exist, better delete, replaced by crowbar, see #4087
```

noraj (changes requested):

Issue #4165, opened 2024-06-22
wkhtmltopdf: outdated and seems no longer maintaining.


astro-angelfish:
- [x] I have searched open and closed issues for duplicates
- [x] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

wkhtmltopdf is no longer working. Invoking the command directly causes `wkhtmltopdf: error while loading shared libraries: libicuuc.so.72: cannot open shared object file: No such file or directory` and I'm unable to build it by myself

Its repository is already archived and its website does not seem to be updated.

### Steps to reproduce

- Just execute `mkhtmltopdf`
- See the error

**Actual result:** Describe here what happens after you run the steps above (i.e. the buggy behaviour)

`wkhtmltopdf: error while loading shared libraries: libicuuc.so.72: cannot open shared object file: No such file or directory`

**Expected result:** Describe here what should happen after you run the steps above (i.e. what would be the correct behaviour)

It should at least print some help information

### Info for developers

**Device:** Hasee TX6-CU5DS with 16GB memory
**GNU/Linux distribution:** Arch Linux
**Tool version:** 0.12.6-2

### Link to debug log

`wkhtmltopdf: error while loading shared libraries: libicuuc.so.72: cannot open shared object file: No such file or directory`

D3vil0p3r:
The issue is caused by `qt5-webkit` dependency because the BA version of this qt5 pkg has been compiled with previous libs. Once this PR has been merged and the mentioned tool rebuilt, it should work

https://github.com/BlackArch/blackarch/pull/4287

PR #4287, opened 2024-10-09
qt5-webkit: version bump

D3vil0p3r:
Fix https://github.com/BlackArch/blackarch/issues/4165

After merging this update, I think that [wkhtmltopdf](https://github.com/BlackArch/blackarch/blob/master/packages/wkhtmltopdf/PKGBUILD) needs to be rebuilt in order to refer to the new libraries.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+qt5-webkit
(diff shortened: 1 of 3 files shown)


noptrix:
@D3vil0p3r thank you

Issue #4209, opened 2024-09-18
Rename valhalla to valhallaapi


derEremit:
My problem.
valhalla from blackarch is trying to replace an aur package valhalla which is an openstreetmap routing engine

Reasoning:

1.
pretty much every package named valhalla in every distro is pointing to the valhalla routing engine
https://repology.org/project/valhalla/versions

2.
the valhalla in blackarch is also called valhallaapi on github
https://github.com/NextronSystems/valhallaAPI

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4289

PR #4289, opened 2024-10-09
valhalla: rename to valhalla-api

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4209

I'm not sure if it is the right process. The `valhalla` package should be removed from the main mirror repo.

### DIFF
--- a/packages/valhalla-api/PKGBUILD
+++ b/packages/valhalla-api/PKGBUILD
@@ -1,7 +1,7 @@
 # This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
 # See COPYING for license details.
 
-pkgname=valhalla
+pkgname=valhalla-api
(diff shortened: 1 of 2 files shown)


packages/valhalla-api/PKGBUILD
```diff
@@ -1,17 +1,17 @@
 # This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
 # See COPYING for license details.
 
-pkgname=valhalla
+pkgname=valhalla-api
```
noraj, line 4: I checked, there is no BA package depending on it, so there is no other package to update.

D3vil0p3r: Good. Let's wait for PR merging.

PR #4292, opened 2024-10-11
add dep for wkhtmltopdf

djustice:
icu72-bin required


noptrix:
@djustice thank you!

noptrix:
@djustice hmm, there is no such package `icu72-bin` from community.

djustice:
correct, it's from aur. but it's required to build. channel said to add it
so i did.

PR #4296, opened 2024-10-14
upgpkg: urlcrazy 0.7.3

djustice:
web page is outdated, their git seems inactive (2019) but still years newer than 0.5...


noptrix:
@djustice thx!

PR #4315, opened 2024-10-30
fix: fix recon-ng with python versions

m7medVision:
recon-ng had problems with python3.12 because it is an old API. I updated PKGBUILD and make it uses python39

### DIFF
--- a/packages/recon-ng/PKGBUILD
+++ b/packages/recon-ng/PKGBUILD
@@ -5,49 +5,51 @@ pkgname=recon-ng
 pkgver=1025.470f4c1
 pkgrel=1
 epoch=1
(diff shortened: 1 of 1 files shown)


noraj (changes requested):
apply a patch and keep python 3.12

m7medVision:
In light of the fact that the last commit on the recon-ng repository was seven months ago, downgrading to Python 3.9 seems like a more stable solution than waiting for upstream changes.

noraj:
It would be best to apply a one-line sed patch and keep the system package style rather that downgrade to an old python version and be forced to use a virtual environment. We keep the virtual environment workaround only when it's unavoidable due to dependencies not incompatibilities or too much dependencies.

PR #4327, opened 2024-11-02
crowbar: fix license and PKGBUILD template base

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4280#issuecomment-2453054670


noraj (approved):
LGTM

PR #4336, opened 2024-11-07
analyzemft

djustice:
python2 build and ver bump


djustice:
so this was accidental... i imagine much of this will need actual review... still learning git apparently.

noptrix:
@djustice hey, did ou re-open this by accident? it has a lot of commits for different packages, which is not the way to go.

djustice:
that's very possible. sorry, i'm re-learning git. i had a stroke a while
back.

PR #4337, opened 2024-11-07
impacket-ba: bump version

xnotsure28x:
impacket-ba: bump version


noraj:
by the way this was not tested, as a bump is not enough, dependencies changes (added and removed), I'm working on it.

Issue #2494, opened 2019-10-14
[Meta] packages with source unavailable


noraj:
packages with source unavailable because upstream URL is dead or redirect to nothing or domain was sold:

- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/blackhash/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/packerid/PKGBUILD --> https://github.com/sooshie/packerid/ ?
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/shellcodecs/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/vncrack/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/metoscan/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/phoss/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/ike-scan/PKGBUILD
- [ ] https://github.com/BlackArch/blackarch/blob/master/packages/dnsmap/PKGBUILD

noraj:
http://16s.us/blackhash/ is dead and I wasn't able to find another source

noraj:
Same fpr https://github.com/BlackArch/blackarch/blob/master/packages/packerid/PKGBUILD

noraj:
Yeah you're right http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.24.0.tar.xz redirects to http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.24.0.tar.xz which doesn't exists anymore.

D3vil0p3r:
**packerid**: https://github.com/sooshie/packerid/ should be right.

**vncrack**:
* URL: `url='http://194.150.168.69/vncrack/'`
* Source: `source=("http://194.150.168.69/vncrack/vncrack_src-1.21.tar.gz")`

**ike-scan**:
* Source: `source=("https://github.com/royhills/ike-scan/archive/refs/tags/1.9.5.tar.gz")`

**dnsmap**:
*URL: `url='https://github.com/resurrecting-open-source-projects/dnsmap'`
* Source: `source=("git+https://github.com/resurrecting-open-source-projects/dnsmap.git")`

What about the other ones? Will they be removed from the repository?

anunna:
Yes they will be removed, but currently we are holding off on any package removals until further notice.

[...]

PR #4338, opened 2024-11-07
ike-scan: fix build error

D3vil0p3r:
Fix build error related to the usage of 64-bit integer.

Close https://github.com/BlackArch/blackarch/issues/2494


noptrix:
@D3vil0p3r thank you!

PR #4347, opened 2024-11-16
Remove duplicated "arch" entries

D3vil0p3r:
Remove "arch" duplicated entries and update dmde.


D3vil0p3r:
@noptrix master branch does not reflect the proposed changes in this merged PR.

Issue #3310, opened 2022-01-14
xray (web scanner)


m0tky:
- **Name**:xray
- **Source URL**:https://github.com/chaitin/xray
- **Description**:https://docs.xray.cool/#/

noraj:
It's a different `xray` than the one requested in #1794 and added in https://github.com/BlackArch/blackarch/commit/700c0725a6aba8119bd0d58123ebd88d7bd72304. The one already included is a recon tool wheras this one is a web scanner.

As the name is conflicting we will have to think about another package name and bin name.

D3vil0p3r:
`xray-toolset` could be well. Xray 2.0 seems to not have anymore a `xray` binary as the old `1.0` but it uses:
* [xpoc](https://docs.xray.cool/tools/xpoc/BasicIntroduction)
* [xapp](https://docs.xray.cool/tools/xapp/BasicIntroduction)
* [xlint](https://docs.xray.cool/tools/xlint/QuickStart)

I don't see any docs related to build from source so I guess that the best approach should be retrieving binaries.

D3vil0p3r:
To be closed: https://github.com/BlackArch/blackarch/pull/4355#issuecomment-2481552633

PR #4355, opened 2024-11-17
xray-toolset: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3310

### DIFF
--- a/packages/xray-toolset/PKGBUILD
+++ b/packages/xray-toolset/PKGBUILD
@@ -1,0 +1,60 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=xray-toolset
+pkgver=EvilPot.0.0.2.r1.gaee9f9c
+pkgrel=1
+pkgdesc='A complete security assessment tool that supports common web security issues scanning and custom POC.'
+arch=('x86_64')
+groups=('blackarch' 'blackarch-scanner' 'blackarch-webapp')
+url='https://github.com/chaitin/xray'
+license=('custom:unknown')
+depends=()
+makedepends=('git')
+source=("git+https://github.com/chaitin/xray.git"
+        "git+https://github.com/chaitin/xray-plugins.git"
+        "https://github.com/chaitin/xray/releases/download/xpoc-0.1.0/xpoc_linux_amd64"
+        "https://github.com/chaitin/xray/releases/download/xapp-0.0.2/xapp_linux_amd64"
+        "https://github.com/chaitin/xray-plugins/releases/download/xlint-0.0.3/xlint_linux_amd64")
+sha512sums=('SKIP'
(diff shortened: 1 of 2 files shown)


packages/xray-toolset/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+pkgdesc='A complete security assessment tool that supports common web security issues scanning and custom POC.'
+arch=('x86_64')
+groups=('blackarch' 'blackarch-scanner' 'blackarch-webapp')
+url='https://github.com/chaitin/xray'
+license=('custom:unknown')
+depends=()
+makedepends=('git')
+source=("git+https://github.com/chaitin/xray.git"
+        "git+https://github.com/chaitin/xray-plugins.git"
+        "https://github.com/chaitin/xray/releases/download/xpoc-0.1.0/xpoc_linux_amd64"
```
noraj, line 16: add xpoc as dependency and built it https://github.com/chaitin/xpoc

packages/xray-toolset/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+        "https://github.com/chaitin/xray/releases/download/xapp-0.0.2/xapp_linux_amd64"
```
noraj, line 17: add xapp as dependency and build it https://github.com/chaitin/xapp

packages/xray-toolset/PKGBUILD
```diff
@@ -0,0 +1,60 @@
...
+        "https://github.com/chaitin/xray-plugins/releases/download/xlint-0.0.3/xlint_linux_amd64")
```
noraj, line 18: add xray-plugins as dependency and built it https://github.com/chaitin/xray-plugins/

noraj (changes requested):
create the packages for the dependencies instead of using pre-build binaries

packages/xray-toolset/PKGBUILD
D3vil0p3r: The `xpoc` repository has not source code to build the binary. Should I create a dedicated package for each one of them and still use the pre-build binaries?

noraj: True. Then if it's closed-source, we would avoid shipping pre-built binaries from untrusted sources.

noraj:
Closed source software (see https://github.com/chaitin/xray/blob/master/README_EN.md):

> Note: xray is not open source, just download the built binary files directly. The repository mainly contains community-contributed POCs. Each xray release will automatically package them.

packages/xray-toolset/PKGBUILD
D3vil0p3r: I agree. At this point I would close this PR. Can you please close https://github.com/BlackArch/blackarch/issues/3310 by explicitly writing the reason you just mentioned?

Issue #3382, opened 2022-04-17
nuclei-templates not used by nuclei package


noraj:
Also right now our package `nuclei-templates` is pretty useless and not needed as a dependency of `nuclei`. Why?
Because since 2.4.0 nuclei is auto-downloading and updating the templates to `~/nuclei-templates` and is not aware of what we deployed at `/usr/share/nuclei-templates`.

```
$ nuclei
                                                                                                                                                                                                                                             
                     __     _                                                                                                                                                                                                                
   ____  __  _______/ /__  (_)                                                                                                                                                                                                               
  / __ \/ / / / ___/ / _ \/ /                                                                                                                                                                                                                
 / / / / /_/ / /__/ /  __/ /                                                                                                                                                                                                                 
/_/ /_/\__,_/\___/_/\___/_/   2.6.3                                                                                                                                                                                                          
                                                                                                                                                                                                                                             
                projectdiscovery.io                                                                                                                                                                                                          
                                                                                                                                                                                                                                             
[WRN] Use with caution. You are responsible for your actions.                                                                                                                                                                                
[WRN] Developers assume no liability and are not responsible for any misuse or damage.                                                                                                                                                       
[INF] nuclei-templates are not installed, installing...                                                                                                                                                                                      
[INF] Successfully downloaded nuclei-templates (v8.9.3) to /home/noraj/nuclei-templates. GoodLuck!
[WRN] Found 2 templates with runtime error (use -validate flag for further examination)
```

So to make the nuclei-templates install system wide taken into account by nuclei we should complexify the install steps :

- Define a `config.yml` file (cf. https://nuclei.projectdiscovery.io/nuclei/get-started/#nuclei-config)
    - Add  `disable-update-check: true` to disable automatic nuclei/templates update check
    - Add `templates: \n  - /usr/share/nuclei-templates`
- Create a wrapper `nuclei -config /usr/share/nuclei/config.yaml "$@"`

So the templates would not be downloaded once per user and making the user home messy.

dualfade:
You are correct. I changed my config to this like other distro's --

![image](https://user-images.githubusercontent.com/2522757/170879421-7516931b-4126-4935-8dae-3503bd55c095.png)

Maybe drop in a new yaml ?
I can take and fix this if so.

D3vil0p3r:
@noraj should it go in [remove list](https://github.com/BlackArch/blackarch/blob/master/lists/to-remove)?

noraj:
[...]

PR #4357, opened 2024-11-17
nuclei: add nuclei-template support

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/3382


D3vil0p3r:
@noraj sorry for mentioning. Can you give a look to this PR so we can lose the related issue ticket once merged? Danke

packages/nuclei/config.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Template Filters
+tags: exposures,cve
+author: geeknik,pikpikcu,dhiyaneshdk
+severity: critical,high,medium
+
+# Template Allowlist
+# 
+# Note: This will take precedence over the .nuclei-ignore file and denylist
+# entries (exclude-tags or exclude-templates list).
+include-tags: dos,fuzz # Tag based inclusion
```
noraj: Could you remove or comment anything that is not `templates` and `disable-update-check`.

noraj: Else I tested and it looks ok.

D3vil0p3r: > Else I tested and it looks ok.

So, is it ok the "as-is" or should I just keep the following?
```
templates:
  - /usr/share/nuclei-templates/
disable-update-check: true
```
and comment the remaining things?

noraj: Just keep the following and comment the rest, those are not default values but examples values, we don't want to induce non default configuration, especially to include dangerous template by default etc.

D3vil0p3r: Fixed

D3vil0p3r:
Can you quick review this and approve it if nothing more is needed?

noraj:
it should be good

noptrix (approved):

Issue #4097, opened 2024-03-13
Bruno


ikstream:
- **Name**: Bruno
- **Source URL**:https://github.com/usebruno/bruno
- **Description**:Opensource IDE For Exploring and Testing Api's (lightweight alternative to postman/insomnia)

Since postman moved to a cloud model, it can be problematic to send private API data to a third company during a penetration test.
It doesn't fully cover postman's features yet, but is on a good way and got a lot of traction.

There is already an [AUR package](https://aur.archlinux.org/packages/bruno) available

noraj:
I know we already have some AUR helpers, etc. but in general we shouldn't start to package all tool that are not directly security related and leave them to the AUR. What do you think?

ikstream:
This tool is very handy while performing exclusively API tests. When all you have is a swagger file and only API in scope. You start with known, valid (spends in the information you have been provided with) requests, which can help revalidate tokens or just send requests from there (through a proxy).

D3vil0p3r:
AUR build fails https://aur.archlinux.org/packages/bruno#comment-993943 Error on [bruno-git](https://aur.archlinux.org/packages/bruno-git) too.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4360

PR #4360, opened 2024-11-18
bruno: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4097

### DIFF
--- a/packages/bruno/PKGBUILD
+++ b/packages/bruno/PKGBUILD
@@ -46,6 +46,8 @@ prepare() {
     _ensure_local_nvm
     cd $pkgname
 
+    nvm install
+
     # https://typicode.github.io/husky/how-to.html#ci-server-and-docker
     export HUSKY=0
 
--- a/packages/bruno/PKGBUILD
+++ b/packages/bruno/PKGBUILD
@@ -26,7 +26,7 @@ makedepends=(
 
 source=(
    "${pkgname}-${pkgver}.tar.gz::https://github.com/usebruno/bruno/archive/v${pkgver}.tar.gz"
-   com.usebruno.app.Bruno.desktop
+   "com.usebruno.app.Bruno.desktop"
 )
 
 sha512sums=('de75210f37907bea641b0f93cd1511b3024e27bbacd892a3de8918b514ca1e1e012fbb553a37095b72f397b3ca87b1f7d5be62c7d7b68616e3b0178dade8facf'
--- a/packages/bruno/PKGBUILD
+++ b/packages/bruno/PKGBUILD
@@ -1,0 +1,100 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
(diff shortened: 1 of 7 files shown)


packages/bruno/PKGBUILD
```diff
@@ -0,0 +1,100 @@
...
+pkgname=bruno
+pkgver=1.34.2
+pkgrel=2
+pkgdesc='Opensource API Client for Exploring and Testing APIs.'
+groups=('blackarch' 'blackarch-webapp')
+arch=('x86_64')
+url="https://www.usebruno.com/"
+license=('MIT')
+_electron=electron31
+depends=(
```
noraj, line 16: I'm not sure but there is maybe a style issue here, it should be horizontal until 80 chars.

packages/bruno/PKGBUILD
```diff
@@ -0,0 +1,100 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+# Initial PKGBUILD from AUR.
+# Old Maintainer: Vianney Bouchaud <aur dot vianney at bouchaud dot org>
+
+pkgname=bruno
+pkgver=1.34.2
```
noraj, line 8: For now, we mostly do git packages instead of release ones, could you adapt from the AUR?

packages/bruno/PKGBUILD
```diff
@@ -0,0 +1,100 @@
...
+)
+
+sha512sums=('de75210f37907bea641b0f93cd1511b3024e27bbacd892a3de8918b514ca1e1e012fbb553a37095b72f397b3ca87b1f7d5be62c7d7b68616e3b0178dade8facf'
+            '0c303dd2a3cb1d1628d25b33d515df4fdaf116298520bed5a251d8320d5f771146a625165ab07fbbf7f4698ebc3bcfc7682a44aaa635d980493c7fc208389884')
+
+_ensure_local_nvm() {
+  # lets be sure we are starting clean
+  which nvm >/dev/null 2>&1 && nvm deactivate && nvm unload
+
+  export NVM_DIR="${srcdir}/${pkgname}-${pkgver}/.nvm"
```
noraj, line 39: IS nvm really needed to build an electron app?

https://wiki.archlinux.org/title/Electron_package_guidelines

noraj (changes requested):

packages/bruno/PKGBUILD
D3vil0p3r: Fixed

packages/bruno/PKGBUILD
D3vil0p3r: Fixed

packages/bruno/PKGBUILD
D3vil0p3r: Not really needed. I removed it. I can leave nvm package to be added in BA, if it could be useful for the future for other stuff.

Issue #4353, opened 2024-11-17
killab33z packages


noraj:
https://codeberg.org/killab33z/blackarch-pkgs/

- [ ] adpeas
- [ ] breads
- [ ] certi
- [ ] extractbitlockerkeys
- [ ] finduncommonshares
- [ ] gpowned
- [ ] ldapconsole
- [ ] ldapmonitor
- [ ] ldapwordlistharvester
- [ ] mssqlrelay
- [ ] netscout
- [ ] pre2k
- [ ] python-aiowmi
- [ ] rdwatool
- [ ] sccmhunter
- [ ] targetedkerberoast

D3vil0p3r:
Tell me if you need some help here. Meanwhile I push for your current license requests (4368, 4369) by other means.

D3vil0p3r:
For https://github.com/BlackArch/blackarch/pull/4368 , the [LICENSE](https://github.com/p0dalirius/ExtractBitlockerKeys/blob/main/LICENSE) file has been created. I think the PR can be updated and merged.

noraj:
> For [#4368](https://github.com/BlackArch/blackarch/pull/4368) , the [LICENSE](https://github.com/p0dalirius/ExtractBitlockerKeys/blob/main/LICENSE) file has been created. I think the PR can be updated and merged.

It was updated 1 month ago already. I updated #4369 and #4410 today.

D3vil0p3r:
Ja. I have an open channel with him by email. In addition to https://github.com/BlackArch/blackarch/pull/4416 and https://github.com/BlackArch/blackarch/pull/4408 do you need additional need from him upstream?

EDIT: he just added also the LICENSE file to the last mentioned repositories. I guess we should be ok now.

D3vil0p3r:
All PR have been merged. I guess this issue can be closed.

PR #4369, opened 2024-11-19
add finduncommonshares

noraj:
related to #4353

waiting for a license https://github.com/p0dalirius/pyFindUncommonShares/issues/31


noraj:
https://github.com/p0dalirius/pyFindUncommonShares/blob/main/LICENSE

only need a pkgver bump

PR #4377, opened 2024-11-23
python-yara-python-dex: add package and replace rednaga

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/pull/3630 and https://github.com/BlackArch/blackarch/issues/3296


noptrix:
@D3vil0p3r thanks a lot :)

Issue #4170, opened 2024-07-04
BBOT 


dorkerdevil:
- **Name**: BBOT
- **Source URL**: https://github.com/blacklanternsecurity/bbot
- **Description**: BBOT (Bighuge BLS OSINT Tool) is a recursive internet scanner inspired by [Spiderfoot](https://github.com/smicallef/spiderfoot), but designed to be faster, more reliable, and friendlier to pentesters, bug bounty hunters, and developers.

Please consider making a Pull Request (PR). Here are some helpful resources:

- AL PKGBUILD guide: https://wiki.archlinux.org/index.php/PKGBUILD
- BA PKGBUILD templates: https://github.com/BlackArch/blackarch-pkgbuilds

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4382

PR #4382, opened 2024-11-24
bbot: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4170


D3vil0p3r:
@noraj sorry again for stalking you. On this, I didn't use the [Python standalone PKGBUILD BA template](https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/python/PKGBUILD-python-standalone) because I got some issues, but I used a different method to use the virtualenv. Let me know if it is ok for you.

noraj:
Why is it on draft?

D3vil0p3r:
Because not working in runtime. Look https://github.com/blacklanternsecurity/bbot/issues/2703#issuecomment-3313195420

noraj:
> I just converted bbot PKGBUILD to a virtual environment approach. I would still keep the other PKGBUILDs because useful for users.

where is the virtualenv version then? also there is no point in keeping this non virtualenv version if it's not working

D3vil0p3r:
> > I just converted bbot PKGBUILD to a virtual environment approach. I would still keep the other PKGBUILDs because useful for users.
> 
> where is the virtualenv version then? also there is no point in keeping this non virtualenv version if it's not working

Ja. It should be in the older commit https://github.com/BlackArch/blackarch/pull/4382/commits/1ed1d037e106c38e676424a06fd3ff7ff01dd3b0#diff-4052d14687365b012298fd93b3ccfe8e99e6e71db7756c806cd9b188ccb49975 (on the left side). I removed it because, don't tell me why (I didn't go deep), if I used the previous virtual env approach, it continues to refer to `/tmp/makepkg` directory.

D3vil0p3r:
> > I just converted bbot PKGBUILD to a virtual environment approach. I would still keep the other PKGBUILDs because useful for users.
> 
> where is the virtualenv version then? also there is no point in keeping this non virtualenv version if it's not working

Now it is fixed. Tested and it seems to work correctly.

PR #4407, opened 2024-11-29
go-exploitdb: initial commit.

s7x:


packages/go-expoitdb/PKGBUILD
```diff
@@ -0,0 +1,43 @@
...
+license=('MIT')
+makedepends=('git' 'go')
+source=("git+https://github.com/vulsio/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
```
noraj, line 20:
```suggestion
pkgver() {
  cd $pkgname

  ( set -o pipefail
    git describe --long --tags --abbrev=7 2>/dev/null |
      sed 's/\([^-]*-g\)/r\1/;s/-/./g' ||
    printf "%s.%s" "$(git rev-list --count HEAD)" \
      "$(git rev-parse --short=7 HEAD)"
  )
}
```

cf. https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git

noraj (changes requested):
Else than that it looks good. Too bad for https://github.com/BlackArch/sploitctl 😁

PR #4420, opened 2024-12-03
pkgup: jeb-arm 5.20

djustice:


noptrix:
@djustice thanks for this. next time, please don't forget to run `updpkgsums` to update the SHA512 sums.

PR #4429, opened 2024-12-09
ghauri: fix package.

s7x:
Hi there,
The ghauri package now needs ua_generator.
I recreated it using the standalone template.
I had to trick a little bit for the execution using `exec $pkgname "\$@"`. The default `exec python /usr/share/ghauri/ghauri.py "\$@"` does not work. Sorry if the way I did is not the correct one but it's the only way I managed to run the tool properly.

Hope it's still valid.

Cheers!

### DIFF
--- a/packages/ghauri/ghauri.install
+++ b/packages/ghauri/ghauri.install
@@ -16,4 +16,3 @@ post_remove() {
   # /usr/share/ghauri/venv
   rm -r /usr/share/ghauri
 }
-


packages/ghauri/PKGBUILD
```diff
@@ -9,27 +9,42 @@ arch=('any')
 groups=('blackarch' 'blackarch-webapp' 'blackarch-exploitation')
 url='https://github.com/r0oth3x49/ghauri'
 license=('MIT')
-depends=('python' 'python-chardet' 'python-tldextract' 'python-colorama'
-         'python-requests')
-makedepends=('git' 'python-setuptools')
+depends=('python')
```
noptrix, line 12: @s7x thx mate :) if there is no version bump then `pkgrel++` i.e. `pkgrel=2` please.

packages/ghauri/ghauri.install
```diff
@@ -0,0 +1,18 @@
...
+}
+
+post_upgrade() {
+  post_install "$@"
+}
+
+post_remove() {
+  # /usr/share/ghauri/venv
+  rm -r /usr/share/ghauri
+}
```
noptrix, line 18: last line should be `\n` only :)

noptrix (changes requested):

noptrix (approved):

PR #4430, opened 2024-12-13
Adding a fix to JDK/JDR as well as links to key issues

xXxSpicyBoiiixXx:


xXxSpicyBoiiixXx:
Adding a script to update the complete system.

xXxSpicyBoiiixXx:
Added a bit more, mirror.rackspace.com is a bit slow so you might need to run it a few times.

xXxSpicyBoiiixXx:
Circling back on this, this will need to be added as without most users will not be able to update their system.

noptrix:
obsolete

Issue #4428, opened 2024-12-08
I think you have a bugged version of smtp-user-enum in your repo.


sudo-ML:
- [ x] I have searched open and closed issues for duplicates
- [x ] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

I ran into an issue on HTB recently because I was not able to set the timeout of smtp-user-enum. I believe the version in the repo is bugged.

The version of smtp-user-enum that is installed seems to be this one: https://github.com/pentestmonkey/smtp-user-enum
There is a problem with this version as the -t argument flag is shared between two different options (target and timeout).
See here: https://github.com/pentestmonkey/smtp-user-enum/issues/4
Running smtp-user-enum --help also confirms this.

```
         -t host  Server host running smtp service
        -T file  File of hostnames running the smtp service
        -p port  TCP port on which smtp service runs (default: 25)
        -d       Debugging output
        -t n     Wait a maximum of n seconds for reply (default: 5)
        -v       Verbose
        -h       This help message
```

This issue was already resolved in 2017 but the pull request was never accepted: https://github.com/pentestmonkey/smtp-user-enum/pull/2

To solve this issue I manually edited the script in /usr/bin/smtp-user-enum to the version in the pull request.

### Steps to reproduce

- Installed smtp-user-enum `sudo pacman -S smtp-user-enum`
- Ran `smtp-user-enum --help`

**Actual result:** You can see that -t is set to two arguments. Target host, and timeout time. 

**Expected result:** The correct behavior is to have -t be target host and -w be timeout time. This is how it works in smtp-user-enum on kali and parrot.

### Info for developers

**GNU/Linux distribution: Arch Linux
**Tool version:**  smtp-user-enum 1.2-6

```options are:
        -m n     Maximum number of processes (default: 5)
        -M mode  Method to use for username guessing EXPN, VRFY or RCPT (default: VRFY)
        -u user  Check if user exists on remote system
        -f addr  MAIL FROM email address.  Used only in "RCPT TO" mode (default: user@example.com)
        -D dom   Domain to append to supplied user list to make email addresses (Default: none)
                 Use this option when you want to guess valid email addresses instead of just usernames
                 e.g. "-D example.com" would guess foo@example.com, bar@example.com, etc.  Instead of 
                      simply the usernames foo and bar.
        -U file  File of usernames to check via smtp service
        -t host  Server host running smtp service
        -T file  File of hostnames running the smtp service
[...]

PR #4431, opened 2024-12-14
smtp-user-enum: patch wrong argument

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4428


noptrix:
@D3vil0p3r thanks a lot

PR #4437, opened 2025-01-01
pyinstaller: update and fix to v6

### LINKED ISSUE #4436: Updating Pyinstaller
Hi,

I'm using Python for creating my own tools. For making it more productive to use them I export my scripts with pyinstaller.
This is not possible on Blackarch, because the version of pyinstaller is very old, has many problems an does not really work with python 3 Syntax. Pip does not work for installing an up to date version and wants me to use pacman, but pacman is finding only the old version in the Blackarch repository.
So I wanted to ask if the pyinstaller could be updated.

noraj:
fix #4436

### DIFF
--- a/packages/pyinstaller/PKGBUILD
+++ b/packages/pyinstaller/PKGBUILD
@@ -9,7 +9,7 @@ groups=('blackarch' 'blackarch-misc')
 pkgdesc='Bundles a Python application and all its dependencies into a single package.'
 arch=('x86_64' 'aarch64')
 url='https://pyinstaller.org'
-license=('custom:mixed')
+license=('Apache-2.0' 'GPL-2.0-or-later' 'MIT')
 depends=(
   'python' 'python-setuptools' 'python-altgraph' # project python dependencies
   'glibc' 'binutils' 'gcc' 'base-devel' # project system dependencies
--- a/packages/pyinstaller/PKGBUILD
+++ b/packages/pyinstaller/PKGBUILD
@@ -8,7 +8,7 @@ epoch=2
 groups=('blackarch' 'blackarch-misc')
 pkgdesc='Bundles a Python application and all its dependencies into a single package.'
 arch=('x86_64' 'aarch64')
-url='http://www.pyinstaller.org/downloads.html'
+url='https://pyinstaller.org'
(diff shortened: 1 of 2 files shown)


packages/pyinstaller/PKGBUILD
```diff
@@ -2,41 +2,72 @@
...
-pkgrel=4
+pkgver=6.11.1
+pkgrel=1
 epoch=2
 groups=('blackarch' 'blackarch-misc')
-pkgdesc='A program that converts (packages) Python programs into stand-alone executables, under Windows, Linux, Mac OS X, Solaris and AIX.'
-arch=('any')
+pkgdesc='Bundles a Python application and all its dependencies into a single package.'
+arch=('x86_64' 'aarch64')
 url='http://www.pyinstaller.org/downloads.html'
```
D3vil0p3r, line 11:
```suggestion
url='https://pyinstaller.org'
```
The `downloads.html` page does not exist anymore.

packages/pyinstaller/PKGBUILD
```diff
@@ -2,41 +2,72 @@
...
-license=('GPL')
-depends=('python2' 'python2-requests' 'python2-pyusb' 'python2-dis3'
-         'python2-pefile' 'python2-macholib' 'python2-setuptools'
-         'python2-altgraph')
-source=("https://files.pythonhosted.org/packages/source/${_pkgname::1}/$_pkgname/$_pkgname-$pkgver.tar.gz")
-sha512sums=('97e7ac3632b5a84a85268c68cb3f3a404ca9f2786396f31016313a15c81c9c453983fd9d0376de0e7543401d3ad8af58b839dde7cefb0dfe7cbb9f3ac360f3e0')
+license=('custom:mixed')
```
D3vil0p3r, line 12:
```suggestion
license=('Apache-2.0' 'GPL-2.0-or-later' 'MIT')
```
Ref.
https://github.com/pyinstaller/pyinstaller/blob/develop/COPYING.txt
https://pyinstaller.org/en/stable/license.html

D3vil0p3r (changes requested):

D3vil0p3r (approved):

PR #4443, opened 2025-01-09
Update cariddi PKGBUILD to include the correct tool category

Qrizzle:
Noticed a little typo when scraping the tools site for tool categories. "blackarch-webap" and "blackarch-webapp" was listed. This little change should fix it.


noptrix:
@Qrizzle thank youvery much

PR #4457, opened 2025-01-31
bkcrack: enhance

Xeonacid:
make build() align with Arch Cmake Package Guidelines

Use doxygen to build doc

Run test

Use SPDX license identifier

optdepends on python for tools

Install example to doc folder

depends on glibc

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+bkcrack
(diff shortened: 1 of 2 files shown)


noptrix:
@Xeonacid thanks a lot - btw, we don't add `check()` and tests... i'll remove it here

Xeonacid:
> btw, we don't add `check()` and tests... i'll remove it here

May I ask why tests weren't included? Having tests enabled helps ensure (to a certain extent) that the package functions as expected.

PR #4458, opened 2025-01-31
rmvpkg: python-rarfiile

Xeonacid:
python-rarfiile seems to be some copy-paste and typo error, which PKGBUILD is python-neotime. It is also not in the repo. The file seems totally useless.


noptrix:
@Xeonacid thank you!

PR #4460, opened 2025-01-31
rmvpkg: log4cxx

Xeonacid:
has been in added in [AL [extra]](https://archlinux.org/packages/extra/x86_64/log4cxx/)


noptrix:
@Xeonacid thank you!

PR #4462, opened 2025-02-07
Add South Korean Mirror

xenix4845:
Hello, we have newly added BlackArch and would like to register for the official mirror. The location is South Korea.

Contact email: mirror@mail.daxhub.net


noptrix:
@xenix4845 thanks, but mirror seems to be down (?)

xenix4845:
> @xenix4845 thanks, but mirror seems to be down (?)

No, the mirror is working fine, but it may not seem to work because it is inaccessible in some countries. 

For example, Kali Linux MirrorBits frequently accesses our mirrors to determine if they are accessible, and if you look here at [Kali Linux](http://cdimage.kali.org/README?mirrorlist) , you can see that our mirror server is working well.

PR #4465, opened 2025-02-12
above: add package.

s7x:
Added [above](https://github.com/casterbyte/Above), a network sniffer.

### DIFF
--- a/packages/above/PKGBUILD
+++ b/packages/above/PKGBUILD
@@ -19,7 +19,12 @@ sha512sums=('SKIP')
 pkgver() {
   cd $pkgname
 
-  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+  ( set -o pipefail
+    git describe --long --tags --abbrev=7 2>/dev/null |
+      sed 's/\([^-]*-g\)/r\1/;s/-/./g' ||
+    printf "%s.%s" "$(git rev-list --count HEAD)" \
+      "$(git rev-parse --short=7 HEAD)"
+  )
 }
 
 build() {


packages/above/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+depends=('python' 'python-virtualenv' 'scapy' 'python-colorama')
+makedepends=('git' 'python-setuptools' 'python-pip')
+source=("$pkgname::git+https://github.com/casterbyte/Above.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  echo $(git rev-list --count HEAD).$(git rev-parse --short HEAD)
+}
```
noraj, line 23:
```suggestion
pkgver() {
  cd $pkgname

  ( set -o pipefail
    git describe --long --tags --abbrev=7 2>/dev/null |
      sed 's/\([^-]*-g\)/r\1/;s/-/./g' ||
    printf "%s.%s" "$(git rev-list --count HEAD)" \
      "$(git rev-parse --short=7 HEAD)"
  )
}
```

ref. https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git

there are git tags https://github.com/casterbyte/Above/tags

packages/above/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgname=above
+pkgver=222.4789d6e
+pkgrel=1
+pkgdesc='Network Protocols Sniffer.'
+arch=('any')
+groups=('blackarch' 'blackarch-sniffer' 'blackarch-scanner'
+        'blackarch-fingerprint' 'blackarch-networking' 'blackarch-recon'
+        'blackarch-voip')
+url='https://github.com/casterbyte/Above'
+license=('GPL3')
```
noraj, line 13: SPDX identifiers please https://spdx.org/licenses/

ref. https://wiki.archlinux.org/title/PKGBUILD#license

packages/above/above.install
```diff
@@ -0,0 +1,17 @@
+post_install() {
+  set -e
+  cd /usr/share/above
+  python -m venv venv
+  source venv/bin/activate &&
+    pip install --isolated --root="/usr/share/above" --prefix='venv' \
```
noraj, line 6: the dependencies look simple, why use a virtual env? https://github.com/casterbyte/Above/blob/4789d6eed0dce7aaf6103852fe7915c6e1d56a5d/setup.py#L16-L18

noraj (changes requested):

Issue #4475, opened 2025-02-26
[tool request] seccomp-tools


astro-angelfish:
- **Name**: seccomp-tools
- **Source URL**: https://github.com/david942j/seccomp-tools
- **Description**: Provide powerful tools for seccomp analysis.

D3vil0p3r:
Enjoy https://github.com/BlackArch/blackarch/pull/4477

astro-angelfish:
> Enjoy [#4477](https://github.com/BlackArch/blackarch/pull/4477)
> 
> Edit: now I see you already opened a PR... I'm closing/drafting mine so you can practice on creating PKGBUILDs here. Your PR has some missing information. Try to study my PR line by line and integrate in yours. Tips: `ruby-os` is a needed dependency that does not exist in official Arch or BA repositories so you need to add it as well.

Well I used `bundle install` from https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/ruby/ruby-standalone.install#L7 and [(ab)used environment variables to use local package installed by `bundler install`](https://github.com/BlackArch/blackarch/blob/87ac28401ef5e15f7fc17789fc60ba4a6b1536ad/packages/seccomp-tools/PKGBUILD#L57) to resolve dependency issues like `ruby-os`.

I've also noticed using their provided binary directly would lead to `LoadError` related errors, so [I crafted this patch](https://github.com/BlackArch/blackarch/blob/87ac28401ef5e15f7fc17789fc60ba4a6b1536ad/packages/seccomp-tools/0001-Add-.-lib-to-library-load-path.patch) to mitigate it.

I didn't know whether creating a patch to customize things like loading path and use environment variable to override library search path, or just creating dependencies package (especially I don't know hou much potential packages I need to create) is better.

D3vil0p3r:
If you want a more compact coding, you could leverage on this template https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/ruby/PKGBUILD-ruby-git so you avoid to use those additional files like the patch and the .install file.

ngn13:
any plans to merge one of the prs?

astro-angelfish:
> any plans to merge one of the prs?

I have no idea. I guess it's permission and responsibility of BA team to choose the best PR.


> If you want a more compact coding, you could leverage on this template https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/ruby/PKGBUILD-ruby-git so you avoid to use those additional files like the patch and the .install file.

[...]

PR #4476, opened 2025-02-28
Tool/seccomp tools

astro-angelfish:
Closes https://github.com/BlackArch/blackarch/issues/4475

It requires rake to compile binary module so I had to append `bundle exec rake` related things in `.install` file, without further code inspection that throws hundreds of warning and generates unwanted stuff

Patches are used to maintain the lib load path while pwd can be everywhere

Disk usage may still need to grind, I noticed multiple examples in spec directory, and still trying to figure out which file is safe to delete.

Sorry for my messy words as my mind was not clear the time when this PR created.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+seccomp-tools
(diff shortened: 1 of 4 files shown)


D3vil0p3r:
@noraj can you close this? Your https://github.com/BlackArch/blackarch/pull/4613 has been merged. This PR is not needed anymore.

PR #4481, opened 2025-03-06
above: add package.

s7x:
Added the [Above](https://github.com/casterbyte/Above) tool by casterbye.

### DIFF
--- a/packages/above/PKGBUILD
+++ b/packages/above/PKGBUILD
@@ -1,0 +1,52 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=above
+pkgver=v2.8.r0.gb73cdec
+pkgrel=1
+pkgdesc='Network Protocols Sniffer.'
+arch=('any')
+groups=('blackarch' 'blackarch-sniffer' 'blackarch-scanner'
+        'blackarch-fingerprint' 'blackarch-networking' 'blackarch-recon'
+        'blackarch-voip')
+url='https://github.com/casterbyte/Above'
+license=('Apache-2.0')
+depends=('python' 'python-colorama' 'scapy')
+makedepends=('python-build' 'python-pip')
+source=("$pkgname::git+https://github.com/casterbyte/Above.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  ( set -o pipefail
+    git describe --long --tags --abbrev=7 2>/dev/null |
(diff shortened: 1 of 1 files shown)


packages/above/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+pkgver=v2.8.r0.gb73cdec
+pkgrel=1
+pkgdesc='Network Protocols Sniffer.'
+arch=('any')
+groups=('blackarch' 'blackarch-sniffer' 'blackarch-scanner'
+        'blackarch-fingerprint' 'blackarch-networking' 'blackarch-recon'
+        'blackarch-voip')
+url='https://github.com/casterbyte/Above'
+license=('Apache-2.0')
+depends=('python' 'python-colorama' 'scapy')
```
noraj, line 14: Are you sure it needs `scapy` (cli) and not `python-scapy` (library)?

noraj:
> Ping @noraj, this is why I did not want to change this package too much. With the new templates we have (using the pip version for example), this is the result:
> 
> image
> 
> image
> 
> image

It's not your fault or the template fault, it's badly packaged upstream.

`/usr/bin/above` is the proper wrapper created by pip that works way better than the dirty 2 lines bash wrapper that `cd /usr/share/package`. But here above is poorly developed, instead of using a standard file architecture, havign separate library files than a CLI consuming the lirary, a proper pyproject.toml and a properly declared entrypoint, it's a dirty 1100 LoC one-file script, packaged with the old setup.py and wrongly declared entrypoint. `/usr/bin/above.py` shouldn't be there but somewhere in `/usr/lib/python3.13/site-packages/above/`. `above_oui_dict.py` is also wrongly deployed under `/usr/lib/python3.13/site-packages/above_oui_dict.py` instead of `/usr/lib/python3.13/site-packages/above/above_oui_dict.py` due to the bad `setup.py`.

noraj:
> `/usr/bin/above.py` shouldn't be there but somewhere in `/usr/lib/python3.13/site-packages/above/`. `above_oui_dict.py` is also wrongly deployed under `/usr/lib/python3.13/site-packages/above_oui_dict.py` instead of `/usr/lib/python3.13/site-packages/above/above_oui_dict.py` due to the bad `setup.py`.

This because it blindly uses `packages=find_packages()` without argument AND the python files are on the root of the directory instead of in a `above` folder. etc. I'll submit a package with prepare() that fix upstream mess.

noraj:
Also you use python/PKGBUILD-python-PEP517-pip that is made for pyproject.toml instead of python/PKGBUILD-python-PEP517-setuptools that is made for sertup.py.

PR #4482, opened 2025-03-06
above: add package

noraj:
replaces #4481, #4465, #4464

cc @s7x


noraj:
I'll be able to remove the patches when https://github.com/casterbyte/Above/pull/11 will be merged.

PR #4514, opened 2025-03-31
netexec: fix build + bump pyinstaller

noraj:
I don't know if it's since this bump https://github.com/BlackArch/blackarch/commit/ea655abb9c6a52d1e573cf45f5c18ab976d6bcc7 (because v1.3.0.r459.gb520fdf9 seemed to work) or if it's this upstream commit https://github.com/Pennyw0rth/NetExec/commit/47bb14680a63a9ff4b7a11be81aac129d5bec6e0 (brings dynamic versioning and empty placeholder version) but netexec BA package is no longer working.

I haven't been able to fix the poetry setup, nor migrate to pipx, nor to a classic venv + pip.

poetry seems no longer working since dynamic-versioning, pipx fails with mysterious error that are better understandable when trying to build with venv + pip (git version command not working, vcs environment not found, all that while copying the .git/, having git as dependency or even directly pip install git+https://… in .install, then mysterious /dev/null, and many other black magic issues). I also tried to copy other method in already existing PKGBUILD. Nothing works.

For example, pipx install mentioned on the [official wiki](https://www.netexec.wiki/getting-started/installation/installation-on-unix), works well on my machine, but is not in a chroot or nspawn container.

All this python virtual environment ecosystem constantly evolving and upstream project continuously keeping to complexity it is not maintainable… The irony is that the use of a virtual environment use in the first place was to avoid the dependency hell of this project (precise version pinned, old version pinned, incompatible version with AL system, git versions, etc. that were also updated very often).

But in this very dark time, I found brightness. I tried harder and after hours of unsuccessfully virtual env packaging, I RTFMed harder the official wiki to discover it's now possible to [build the whole python project as a static binary](https://www.netexec.wiki/getting-started/installation/manually-building-the-binary). So I did so and it works! ~~Humanity~~ pentest community is saved!


noraj:
cc @Fireplop

noptrix:
@noraj thank you very much

PR #4531, opened 2025-05-26
spectrwm: patch PKGBUILD and add baraction.sh for custom bar

bubb13gummy:


anunna:
@bunnyhop-dev 
1. The package you are requesting changes on is the base spectrwm package. This was removed from the official Archlinux repos and we then packaged it here. This should not be changed. 
2. We have another PKGBUILD for customization `blackarch-config-spectrwm` which points to a different repo [blackarch-config-spectrwm](https://github.com/BlackArch/blackarch-config-spectrwm). This is where all the custom spectrwm configs are located. 

I am closing this PR. Open a new PR on [blackarch-config-spectrwm](https://github.com/BlackArch/blackarch-config-spectrwm). Please avoid unnecessary dependencies for bars since our spectrwm bars have no deps. If there are must have/life changing deps for your bar, please make sure to update the [PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/blackarch-config-spectrwm/PKGBUILD) for the spectrwm config

PR #4542, opened 2025-06-17
arpspoof-smikims: add package

D3vil0p3r:


noptrix:
@D3vil0p3r i think this will conflict with `dsniff`'s `arpspoof` package... could you rename this package and tool, please? maybe sth like `$username-arpspoof` or whatever you prefer

D3vil0p3r:
done

packages/arpspoof-smikims/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+
+  make
+}
+
+package() {
+  cd $_pkgname
+
+  install -Dm 755 arpspoof "$pkgdir/usr/bin/$pkgname"
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md
+}
```
noraj, line 40: If you test with ba-dev and with pkgcheck installed, you'll see the style warning that the final newline is missing.

noraj (changes requested):

packages/arpspoof-smikims/PKGBUILD
noraj: Fixed by noptrix in https://github.com/BlackArch/blackarch/commit/9223540e00ed3b88db64c84df0bb0aece28947f9

PR #4560, opened 2025-07-11
fix: fix the application file for johnny

1p22geo:
PKGBUILD installs the executable at /usr/bin, not /opt/johnny.
And it needs to install the icon too.


noptrix:
@1p22geo thank you

PR #4562, opened 2025-07-14
peass: fix wrong version

D3vil0p3r:
The current `peass` PKGBUILD had erroneous `_pkgver=20221225` that retrieved very old  linpeass binaries.

Furthermore, the current build failed because of changes on linpeass builder that requires `--output` parameter.

The proposed PKGBUILD change manages these issues.


noptrix:
@D3vil0p3r thanks a lot.

D3vil0p3r:
Sorry. I needed to do a small fix https://github.com/BlackArch/blackarch/pull/4563

PR #4563, opened 2025-07-14
peass: fix binaries args

D3vil0p3r:
Sorry, in my last PR I mis-inverted building args to generate linpeas binaries.


noptrix:
@D3vil0p3r thank you

PR #4570, opened 2025-07-22
Fix(getsploit, barq): Remove unresolveable python-clint dependency by…

Fig12:
When running sudo pacman -S blackarch, the system fails to install the blackarch-exploitation modules since getsploit and barq both require python-clint. This dependency is unresolvable since it is not available as a standard Pacman package. My solution modifies PKGBUILD for both of those tools by installing clint using pip instead. I have tested it locally and both tools have had all dependencies install completely with my fix.

### DIFF
--- a/packages/barq/PKGBUILD
+++ b/packages/barq/PKGBUILD
@@ -1,6 +1,51 @@
 # This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
 # See COPYING for license details.
 
+pkgname=getsploit
+pkgver=37.bcab2ee
+pkgrel=5
+pkgdesc='Command line utility for searching and downloading exploits.'
+groups=('blackarch' 'blackarch-exploitation' 'blackarch-misc')
+arch=('any')
+url='https://github.com/vulnersCom/getsploit'
+license=('LGPL3')
+depends=('python')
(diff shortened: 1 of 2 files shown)


Sidicer:
Another solution would be to have `python-clint` part of the packages in blackarch repo so that we would not need to track  any time a packges requires it as a dependency :)

packages/barq/PKGBUILD
```diff
@@ -1,6 +1,51 @@
...
+
+[taceo@thinkpad getsploit]$ d ..
+bash: d: command not found
+[taceo@thinkpad getsploit]$ cd ..
+[taceo@thinkpad packages]$ cd barq
+[taceo@thinkpad barq]$ nano PKGBUILD 
+[taceo@thinkpad barq]$ cat PKGBUILD 
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
```
noraj, line 48: You pasted your terminal output here

noraj (changes requested):
I don't think we'll go the pip way. We'll rather package python-clint

packages/barq/PKGBUILD
Fig12: oops, apologies.

PR #4573, opened 2025-07-23
add python-clint

### LINKED ISSUE #4572: create python-clint as the dependency is missing for several tools and not in AL
cf #4570

noraj:
fixes #4572

replaces #4570

tested working


ikstream:
Shouldn’t the pkgrel be set to 16 to avoid issues on systems that have the now dropped package installed.

noraj:
> Shouldn’t the pkgrel be set to 16 to avoid issues on systems that have the now dropped package installed.

you're probably right, so that was a package that was dropped from AL

PR #4579, opened 2025-07-27
Add blackarch-helper package

x0rgus:
- **Name**: blackarch-helper
- **Source URL**: https://github.com/x0rgus/blackarch-helper
- **Description**: A lightweight Bash TUI/CLI to explore, search and install tools from the BlackArch repository.

This tool provides an interactive fzf-powered TUI interface as well as traditional CLI subcommands. It replaces the now deprecated `blackman` tool, which is still referenced in official documentation but no longer maintained or recommended.
s2
s1

License: MIT

### DIFF
--- a/packages/blackarch-helper/.SRCINFO
+++ b/packages/blackarch-helper/.SRCINFO
@@ -1,0 +1,22 @@
+pkgbase = blackarch-helper
+	pkgdesc = TUI/CLI helper to manage and install tools from the BlackArch repository
+	pkgver = 1.1.0
+	pkgrel = 1
+	url = https://github.com/x0rgus/blackarch-helper
+	arch = any
+	groups = blackarch
+	license = MIT
+	depends = bash
+	depends = fzf
+	source = blackarch-helper
+	source = LICENSE
+	source = blackarch-helper.1
+	source = blackarch-helper.bash-completion
+	source = _blackarch-helper
+	sha512sums = SKIP
+	sha512sums = SKIP
+	sha512sums = SKIP
+	sha512sums = SKIP
+	sha512sums = SKIP
+
+pkgname = blackarch-helper
--- a/packages/blackarch-helper/blackarch-helper
+++ b/packages/blackarch-helper/blackarch-helper
@@ -1,0 +1,366 @@
+#!/usr/bin/env bash
+#
+# blackarch-helper
+# Simple CLI helper for BlackArch tools
+# Author: Lucas "x0rgus"
+# License: MIT
+# Repository: https://github.com/x0rgus/blackarch-helper
+#
+# --- Configuration ---
+# Set VERBOSE to "false" by default. It can be overridden by a command-line flag.
+VERBOSE=false
+
+## --- Color and Symbol Configuration ---
+COLOR_RESET=$'\033[0m'
+COLOR_GREEN=$'\033[0;32m'
+COLOR_YELLOW=$'\033[0;33m'
+COLOR_RED=$'\033[0;31m'
+COLOR_CYAN=$'\033[0;36m'
+COLOR_BOLD=$'\033[1m'
+
+# --- Logging Functions ---
+# Helper functions to standardize message output.
+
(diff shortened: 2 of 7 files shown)


noptrix:
@noraj @dualfade @ikstream mind giving this a quick look? seems like it has potential.

ikstream:
I apriciate every contribution and I would really love to see blackam making a comeback. I did only a quick review, so take my feedback with that grain of salt. If I did oversee something, please point it out.

This tool, at it's current state is a wrapper around pacman. The key selling point of blackman was to build packages from source from our repo or basically any repo based on it's pkgbuild file. It had/has it's flaws, that's why it's no longer in use. This tool allows searches through pacman, installs through pacman and list categories through pacman. 

I don't see the blackman replacement as it is mentioned in the description. 
My main Problem with that is, that it adds complexity. Currently if user complain, that something isn't working, it's either our broken keyring, or something with the packages (either ours, or upstream). If user start using install commands through something that goes more in the direction of pamac (Manjaro) we start adding overhead to troubleshoot where the issue lies without adding to much value (from my point of view) to pacman. 
I would prefer user to be able to use pacman properly, especially as most of the commands are provided on the website or the guides. I am all in for enabling user to get competent and learn new stuff. But in my opinion this wouldn't be the right way to it. 

As I mentioned before, I did a quick review, so if anything is wrong, please point it out. I really appreciate the efffort you put in here, with even a man file and autocomplete commands. But from my point of view it's unfortunately in the wrong tool/direction.

noraj:
I agree with @ikstream. In short, yes for the info about packages and search TUI, no to the install package wrapper (note it can lead to [partial upgrade](https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported)).

x0rgus:
thank you for the feedback @ikstream @noraj , taking that into account, i have rewritten the tool from the ground up. It no longer functions as a pacman wrapper and instead build packages from source now. The main features, including the `fzf` based interactive menu to sync, browse categories, and build tools, are now fully operational. I believe this new approach is much more aligned with the project's goals and would appreciate your thoughts on this new version, i'm still working on some features like installing multiple packages, but the core features were migrated.

packages/blackarch-helper/.SRCINFO
```diff
@@ -0,0 +1,22 @@
+pkgbase = blackarch-helper
```
noraj: submitting `.SRCINFO` is only for AUR, we don't need that here

packages/blackarch-helper/blackarch-helper
```diff
@@ -0,0 +1,516 @@
+#!/usr/bin/env bash
```
noraj: Please don't copy all source files locally, deploy them from git. See this template:  https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git

noraj (changes requested):

packages/blackarch-helper/.SRCINFO
x0rgus: done, [commit](https://github.com/BlackArch/blackarch/pull/4579/commits/035c6b7c5ffd082016ab72a99aa4b0944301c78a)

packages/blackarch-helper/blackarch-helper
x0rgus: done, [commit](https://github.com/BlackArch/blackarch/pull/4579/commits/9d7d936ebe3525768e8e8bc197ae25a841ba6582)

noraj: PKGBUILD is better, and you removed SRCINFO, but all the local files are still here, please remove LICENSE, _blackarch-helper, etc. anything but the PKGBUILD. All other files will be fetched from git at build time

x0rgus: removed local files, [commit](https://github.com/BlackArch/blackarch/pull/4579/commits/76bbcbb567f44556c7f3a4d798a871a8d99d4ac2)

PR #4584, opened 2025-08-01
fixed pkgbuild syntax errors

m1sh4k:
I have checked each of `PKGBUILD` files with `bash -n` and found some syntax errors (all of them looked like typos :upside_down_face:) and corrected them.


packages/smikims-arpspoof/PKGBUILD
```diff
@@ -33,6 +33,6 @@ package() {
   install -Dm 755 arpspoof "$pkgdir/usr/bin/$pkgname"
   install -m 755 sweep.sh "$pkgdir/usr/bin/$pkgname-sweep"
   install -m 755 cage.sh "$pkgdir/usr/bin/$pkgname-cage"
-  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md"
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/README.md"
```
noraj: Shouldn't it rather be?

```suggestion
  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
```

Similar to https://github.com/BlackArch/blackarch-pkgbuilds/blob/d44f5dee7e2f20161bdb10199cce7f5a322f4e38/PKGBUILD-generic-jointscript#L37, deploy a file in a folder.

packages/python2-pip/PKGBUILD
```diff
@@ -33,7 +33,7 @@ prepare() {
   # Remove certifi usage
   sed -i 's|from pip._vendor.certifi import where|where = lambda: "/etc/ssl/certs/ca-certificates.crt"|' src/pip/_internal/commands/debug.py
 
-  rm -rf src/pip/_vendor/!(__init__.py)
+  rm -rf 'src/pip/_vendor/!(__init__.py)'
```
noraj: I'm not sure about this one.

packages/smikims-arpspoof/PKGBUILD
m1sh4k: Yep, ur right

packages/python2-pip/PKGBUILD
m1sh4k: Sry, putting quotes was my bad, this file was ok. I should have tested it more carefully. Bash was throwing error because it does not initialize `shopt` with `-n` key.

noptrix:
@m1sh4k thank you

PR #4588, opened 2025-08-05
ci: update ubuntu runner version and actions/checkout version

SherkeyXD:
The runner `ubuntu-20.04` has been deprecated since 2025-04-15, see https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,38 +1,48 @@
 name: CI
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   build:
-     runs-on: ubuntu-latest
-     steps:
-       - uses: actions/checkout@v3
-       - name: Build Package
-         run: |
-           CHANGED_FILES=($(git diff --name-only --diff-filter=d $THE_LAST_COMMIT HEAD))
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -1,38 +1,48 @@
...
+          for FILE in $CHANGED_FILES; do
+             if [[ $(basename $FILE) == "PKGBUILD" ]]; then
+               docker build -t builder -f travis/Dockerfile $(dirname $FILE)
+               docker run --rm builder
+             fi
+          done
+
   pkgcheck:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
noraj: I'm not a big fan of using `latest` tag has it can introduce breaking change silently, I would prefer to use `ubuntu-24.04` instead and manually update it when newer version are released.

ikstream: I strongly agree here

SherkeyXD: shoule we change the runner for task `build` as well?

https://github.com/BlackArch/blackarch/blob/469eb06b720196bb64f4734f4f5ecba11fd99ba9/.github/workflows/tests.yml#L10

noraj: Yes, it does not make sens to have inconsistent values

PR #4593, opened 2025-08-12
hostapd-mana: add package

noraj:
To test, one need to write a `/etc/hostapd/hostapd.conf`. To improve this package maybe we need to provide an example one. See https://github.com/sensepost/hostapd-mana/wiki/Simplest-hostapd.conf.

Also, there is maybe a way to make the mana fork not conflicting with original hostapd, but I didn't want to spend time on that.

### DIFF
--- a/packages/hostapd-mana/PKGBUILD
+++ b/packages/hostapd-mana/PKGBUILD
@@ -10,14 +10,17 @@ groups=('blackarch' 'blackarch-wireless')
 url='https://github.com/sensepost/hostapd-mana'
 license=('custom:BSD/GPL v2 dual')
 depends=('openssl' 'libnl')
+optdepends=('asleap')
 makedepends=('git')
 source=("git+https://github.com/sensepost/$pkgname.git"
-        'hostapd.service')
+        "$pkgname.service"
+        "$pkgname@.service"
+        "$pkgname.tmpfiles")
 sha512sums=('SKIP'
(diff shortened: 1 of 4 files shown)


OscarAkaElvis:
Usually this is invoked as hostapd-mana instead of hostapd. At least is the way it is done in other Linux distributions. Maybe renaming the binary or creating a symbolic link is enough to avoid the possible conflict.

noptrix:
@noraj i think we shall `strcat()` `-mana` to bins and config dirs+files.

noraj:
@OscarAkaElvis @noptrix It should be good with binaries, pid, services files, directories, config values, etc. updated.

OscarAkaElvis:
Looks good except for one thing. What is this for? `optdepends=('asleap')` . I think asleap is not a dependency of this package AFAIK.

noraj:
> Looks good except for one thing. What is this for? `optdepends=('asleap')` . I think asleap is not a dependency of this package AFAIK.

It is for crackapd (part of hostapd-mana): https://github.com/sensepost/hostapd-mana/blob/8853d5ac7f97b140b7899d2cf074eca921e55c3b/crackapd/crackapd.conf#L38

And anyway it's optional dependency, that is not installed by default. It's hardly just a suggestion for users.

OscarAkaElvis:
Great then, go ahead with the package please. Thank you!

PR #4609, opened 2025-09-17
blackarch-devtools-toolkit: added needed dependency

D3vil0p3r:
This fix prevents the `btrfs command not found` error in case `btrfs-progs` package is not installed.


packages/blackarch-devtools-toolkit/PKGBUILD
```diff
@@ -30,4 +36,3 @@ package() {
 
   install -Dm 755 "target/release/$_pkgname" "$pkgdir/usr/bin/ba-dev"
 }
-
```
noraj, line 33: Don't remove the trailing blank line. see https://github.com/BlackArch/blackarch/blob/master/.editorconfig. If it was removed automatically by your editor, either use a [EditorConfig capable](https://editorconfig.org/#pre-installed) editor or install [the plugin](https://editorconfig.org/#download) corresponding to your current editor.

D3vil0p3r: Sorry, yes. Often I destroy and rebuilt my systems, sometimes I forget to configure the editor. Now fixed.

Issue #4543, opened 2025-06-17
bloodhound: gconf dependency is still needed?


D3vil0p3r:
Type: Question

## Question

Is `gconf` dependency still needed for [BloodHound-Legacy](https://github.com/BlackArch/blackarch/blob/master/packages/bloodhound/PKGBUILD)?

`gconf` carries `python2` dependency.

saltasatelites:
Hi
BloodHound-Legacy has been deprecated.
Use instead BloodHound Community Edition  https://github.com/SpecterOps/BloodHound

D3vil0p3r:
I don't think. Look the commit history.

saltasatelites:
Take a look to the project. https://github.com/SpecterOps/BloodHound-Legacy?tab=readme-ov-file#bloodhound-legacy-edition-v4-has-been-deprecated

D3vil0p3r:
That statements says:
```
This repository is for BloodHound Legacy (version 4), which was last updated in 2023 and is no longer maintained. It will be archived in the near future.
```
The last update of the repository is 2 days ago. This point can be clarified directly by asking to the maintainer.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4612

PR #4612, opened 2025-09-18
bloodhound: removing unused dependencies

D3vil0p3r:
Remove `alsa-lib gtk3 libxss nss gconf libxtst` dependencies. No references found in the source code (unless I'm drunk).

Close: https://github.com/BlackArch/blackarch/issues/4543

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+bloodhound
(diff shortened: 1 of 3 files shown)


D3vil0p3r:
@noraj I removed unused dependencies (`gconf` also depends on Python2). I don't see their usage on the [source code](https://github.com/SpecterOps/BloodHound-Legacy). I also checked the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bloodhound) and they are not used there. Just to do a double check, can you give a look you too please?

Issue #4547, opened 2025-06-18
grpc-pentest-suite


vdun:
- **Name**: grpc-pentest-suite
- **Source URL**: https://github.com/nxenon/grpc-pentest-suite
- **Description**: gRPC-Web Pentesting Suite

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4618

PR #4618, opened 2025-09-19
grpc-pentest-suite: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4547


packages/grpc-pentest-suite/grpc-pentest-suite.install
```diff
@@ -0,0 +1,16 @@
+post_install() {
```
noraj: did you try without virtualenv?

dependencies are all present in AL repo and there are not a lot of them https://github.com/nxenon/grpc-pentest-suite/blob/main/requirements.txt

virtualenv is not the default, it's the last resort when normal packaging is not working due to requiring non latest dependencies or when there a 50 dependencies with frequent update. When everything is present in AL repo and there are only a few, virtualenv must be avoided. It generates way more issues and weight way more space.

D3vil0p3r: Fixed

noraj (approved):
LGTM

Issue #4557, opened 2025-07-08
python-ldapdomaindump (AL) vs ldapdomaindump (BA)


noraj:
```
(291/291) analyse des conflits entre fichiers                                                                                   [##############################################################################] 100%
erreur : la validation de la transaction a échoué (conflit de fichiers )                                                                                                                                             
python-ldapdomaindump : /usr/bin/ldapdomaindump est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)                                                                                          
python-ldapdomaindump : /usr/bin/ldd2bloodhound est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)                                                                                          
python-ldapdomaindump : /usr/bin/ldd2pretty est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)                                                                                              
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__init__.py est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)                                                     
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__main__.py est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__pycache__/__init__.cpython-313.opt-1.pyc est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__pycache__/__init__.cpython-313.pyc est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__pycache__/__main__.cpython-313.opt-1.pyc est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__pycache__/__main__.cpython-313.pyc est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
python-ldapdomaindump : /usr/lib/python3.13/site-packages/ldapdomaindump/__pycache__/convert.cpython-313.opt-1.pyc est déjà présent dans le système de fichiers (appartenant à ldapdomaindump)
[...]

PR #4619, opened 2025-09-19
ldapdomaindump: to remove

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4557

### DIFF
--- a/lists/python2-tools
+++ b/lists/python2-tools
@@ -243,7 +243,6 @@ kwetza
 laf
 lans
 ldap-brute
-ldapdomaindump
 leroy-jenkins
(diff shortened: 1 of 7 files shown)


noraj:
If removing the BA package, it would also require changing the dependency in the packages using it.

noraj (approved):
LGTM

D3vil0p3r:
> If removing the BA package, it would also require changing the dependency in the packages using it.

I already did... unless I forgot something...

noraj:
> I already did... unless I forgot something...

Yes yes, all good. I made a comment after reading the title and before reviewing the changes.

Issue #4615, opened 2025-09-19
myjwt: package broken


dualfade:
myjwt is broken --
local/myjwt 202.719b939-5 (blackarch blackarch-exploitation blackarch-cracker)

testing --
exploit % myjwt eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.KMUFsIDTnFmyG3nMiGM6H9FNFUROf3wh7SmqJp-QV30
/usr/lib/python3.13/site-packages/click/core.py:1193: UserWarning: The parameter -c is used more than once. Remove its duplicate as parameters should be unique.
  parser = self.make_parser(ctx)
/usr/lib/python3.13/site-packages/click/core.py:1186: UserWarning: The parameter -c is used more than once. Remove its duplicate as parameters should be unique.
  self.parse_args(ctx, args)

Your jwt is:
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.KMUFsIDTnFmyG3nMiGM6H9FNFUROf3wh7SmqJp-QV30

Your jwt decoded is:

Header:
alg = HS256
typ = JWT

Payload:
sub = 1234567890
name = John Doe
admin = True
iat = 1516239022

Signature:
"KMUFsIDTnFmyG3nMiGM6H9FNFUROf3wh7SmqJp-QV30"

Traceback (most recent call last):
  File "/usr/bin/myjwt", line 33, in <module>
    sys.exit(load_entry_point('myjwt==2.1.0', 'console_scripts', 'myjwt')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
[...]

PR #4622, opened 2025-09-19
myjwt: fix questionary dependency error

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4615


noptrix:
@D3vil0p3r thank you

Issue #4479, opened 2025-03-03
firebaseEnum


hackeyware:
- **Name**:firebaseEnum
- **Source URL**:https://github.com/Sambal0x/firebaseEnum.git
- **Description**:Tool to mass analyse potentially exposed Firebase databases on Android apps

noraj:
On one hand the tool has not been maintained for 6 years so it violates the [acceptance rules](https://github.com/BlackArch/blackarch/issues/2842) but on the other hand the only dependency is bs4 so if it works with https://archlinux.org/packages/extra/any/python-beautifulsoup4/ that shouldn't pose any problem.

D3vil0p3r:
The tool has no LICENSE. And also produces the annoying warning message:
```
SyntaxWarning: invalid escape sequence '\ '
```
I would close this issue. If the requestor will find a maintained tool with the same purpose, it would be better.

noraj:
I'll let that on the list since I don't see other firebase tools already included.

D3vil0p3r:
In case, I created a PR for it https://github.com/BlackArch/blackarch/pull/4623

PR #4623, opened 2025-09-19
firebaseEnum: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4479


noraj:
Was it not working with classic packaging? There is only bs4 dependency…

D3vil0p3r:
> Was it not working with classic packaging? There is only bs4 dependency…

Done. It also requires `python-requests-futures` otherwise you get:
```
You'll need to pip install requests-futures for this tool.
```
Now should be ok.

packages/firebaseEnum/firebaseEnum.install
```diff
@@ -0,0 +1,16 @@
+post_install() {
```
noraj: Remove the .install too

D3vil0p3r: Ah yes. Done.

Issue #2667, opened 2020-02-26
[Tool request] mercury


vdun:
https://github.com/cisco/mercury

Tazmain:
@vdun Does this request include pmercury as well?

vdun:
Maybe make it in two packages ?
- mercury
- python-pmercury

D4imyo:
This issue should be closed.
i couldn't install the python implementation of this tool because of this [issue](https://github.com/pynetwork/pypcap/issues/105)

D3vil0p3r:
Furthermore, the README says:
```
In the root directory, edit mercury.cfg with the network interface you want to capture from, then run

./configure
make

sudo make install MERCURY_CFG=mercury.cfg
```
so how can we manage  the network interface name inside this config file at packaging time?

noraj:
We'll see what they answer, but it's only possible a compile time and not at runtime then it won't be possible to package it. See https://github.com/cisco/mercury/issues/29#issuecomment-2392248380

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4625

PR #4625, opened 2025-09-22
mercury: fix building errors

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/2667

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,1 @@
+mercury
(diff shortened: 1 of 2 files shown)


D3vil0p3r:
@noraj can you give a look to this? mercury already existed in BA repo since 2020 but its build process was failing. Now it should be fixed.

Issue #4546, opened 2025-06-18
fiked: need to be rebuilt due to error while loading shared libraries: libnet.so.1


D3vil0p3r:
- [X] I have searched open and closed issues for duplicates
- [X] I am submitting a bug report for existing functionality that does not work as intended

----------------------------------------

### Bug description

`fiked` needs to be rebuilt due to the following error:
```
fiked: error while loading shared libraries: libnet.so.1: cannot open shared object file: No such file or directory
```
Currently `libnet` provides `libnet.so.9`.

D3vil0p3r:
Hey @noptrix Kannst du bitte die genannten Pakete neu bauen?

noptrix:
> Hey [@noptrix](https://github.com/noptrix) Kannst du bitte die genannten Pakete neu bauen?

i’ll try rebuilding `fiked firewalk inviteflood rtpbreak heartleech rcracki-mt sipcrack sslsniff thc-ssl-dos xplico`, but as far as i recall several of them couldn’t be rebuilt in the past... they need in-depth code patches to work with recent gcc/lib* versions.

noptrix:
> > Hey [@noptrix](https://github.com/noptrix) Kannst du bitte die genannten Pakete neu bauen?
> 
> i’ll try rebuilding `fiked firewalk inviteflood rtpbreak heartleech rcracki-mt sipcrack sslsniff thc-ssl-dos xplico`, but as far as i recall several of them couldn’t be rebuilt in the past... they need in-depth code patches to work with recent gcc/lib* versions.

only `inviteflood` and `thc-ssl-dos` were rebuild. the other ones need to be patched

D3vil0p3r:
> only `inviteflood` and `thc-ssl-dos` were rebuild. the other ones need to be patched

Thanks. I will try to give a look to the ones to patch. Currently I fixed `fiked`: https://github.com/BlackArch/blackarch/pull/4631

`inviteflood` still returns `inviteflood: error while loading shared libraries: libnet.so.1: cannot open shared object file: No such file or directory`. I checked its [PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/inviteflood/PKGBUILD). Files are retrieved already built... So rebuilding the package is not enough. It is necessary to retrieve the source code of it and build it in the PKGBUILD. I fixed it here https://github.com/BlackArch/blackarch/pull/4633

EDIT:
fixed firewalk https://github.com/BlackArch/blackarch/pull/4632
[...]

PR #4632, opened 2025-09-30
firewalk: fix building errors

D3vil0p3r:
Ref https://github.com/BlackArch/blackarch/issues/4546


noraj (commented):
Looks ok but I'll let someone else validate

PR #4645, opened 2025-10-05
Polished README structure and style as per feedback

Pushpenderrathore:
I’ve simplified the README to make it clean and easy to follow. Let me know if there’s anything else you’d like me to adjust!


noptrix:
@Pushpenderrathore There are still some incorrect heading levels. Please follow the correct structure: # foo for sections, ## bar for subsections, ### for subsubsections, etc. In the changes you made, for example, I see ## Get Involved, but it should be level 1 (# Get Involved).

Please also avoid long descriptions in the section titles.

Also, could you update the table of contents (ToC) accordingly and move it to the beginning of the README?

Please use `set tw=80` and make a line break at 79th character... I mean I would fix it in worst case, but, we're also very busy here. :)

Thank you for your effort!

noptrix:
@Pushpenderrathore thank you very much. :)

Pushpenderrathore:
It's my pleasure. thank you because I learn something from this practical approach.

Issue #4648, opened 2025-10-09
RustHound CE


noraj:
- **Name**: RustHound-CE
- **Source URL**: https://github.com/g0h4n/RustHound-CE

We already have RustHound (https://github.com/NH-RED-TEAM/RustHound) collector for BH v4, we need RustHound-CE, collector for BHv5.

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4649

PR #4649, opened 2025-10-11
rusthound-ce: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4648

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,1 +1,3 @@
 rusthound-ce
+rusthound
+


packages/rusthound-ce/PKGBUILD
```diff
@@ -0,0 +1,43 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=rusthound-ce
+pkgver=55.44b4613
+pkgrel=1
+pkgdesc='Active Directory data collector for BloodHound Community Edition.'
```
noraj:
```suggestion
pkgdesc='Active Directory data collector for BloodHound community edition (v5).'
```

Let's keep the same format as in https://github.com/BlackArch/blackarch/blob/master/packages/bloodhound-python-ce/PKGBUILD and https://github.com/BlackArch/blackarch/blob/master/packages/bloodhound-python/PKGBUILD

noraj: And by the way if you can modify https://github.com/BlackArch/blackarch/blob/master/packages/rusthound/PKGBUILD too.

```diff
-pkgdesc='Active Directory data collector for BloodHound.'
+pkgdesc='Active Directory data collector for Bloodhound legcacy (v4).'
```

PR #4652, opened 2025-10-12
autowordlists: add package

D3vil0p3r:


packages/autowordlists/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=autowordlists
+pkgver=882.6085809
+pkgrel=1
+groups=('blackarch' 'blackarch-fuzzer')
+pkgdesc='Collection of wordlists generated automatically.'
+url='https://github.com/carlospolop/Auto_Wordlists'
```
noraj, line 9: Those lists are just concatenations of list already included (mostly) in blackarch, see https://github.com/carlospolop/Auto_Wordlists/blob/main/scripts/yaml_lists/lists.yaml

- SecLists https://github.com/BlackArch/blackarch/blob/master/packages/seclists/PKGBUILD
- PATT https://github.com/BlackArch/blackarch/blob/master/packages/payloadsallthethings/PKGBUILD
- dirsearch https://github.com/BlackArch/blackarch/blob/master/packages/dirsearch/PKGBUILD


Maybe just package those not already included like:

- https://github.com/1N3/IntruderPayloads

And maybe not included lists like https://github.com/payloadbox as it seems those are just outdated copy of the intruders files of PATT. Eg. https://github.com/payloadbox/ssti-payloads/blob/master/Intruder/ssti-payloads.txt is a 3 years old copy of the more up to date https://github.com/swisskyrepo/PayloadsAllTheThings/blob/master/Server%20Side%20Template%20Injection/Intruder/ssti.fuzz.

D3vil0p3r:
As mentioned above, probably not worth to have.

noraj:
> As mentioned above, probably not worth to have.

So autowordlists will just generate duplicate content, and as it's auto-regenerated every few hours and weight a lot, it will make a huge several hundreds of MB updated every 8 hours. So I'm very against packaging this.

It's like the password leak game, there are few leak sources and a lot of people crafting dumb combo list to get attention or money out of no effort.

PR #4653, opened 2025-10-12
security-wordlist: add package

D3vil0p3r:


packages/security-wordlist/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=security-wordlist
+pkgver=23.a090dff
+pkgrel=1
+groups=('blackarch' 'blackarch-fuzzer')
+pkgdesc='A wordlist repository with human-curated and reviewed content.'
+url='https://github.com/DragonJAR/Security-Wordlist'
```
noraj, line 9: Has it any interest compared to what is already in https://github.com/danielmiessler/SecLists?

https://github.com/BlackArch/blackarch/blob/master/packages/seclists/PKGBUILD

D3vil0p3r: By doing several HTB machines, I saw that the wordlist had several things that I didn't find in seclists... probably it was just lucky to have the right strings for those HTB machines.

D3vil0p3r: I thought about your point. Not worth to upload it. Closing.

noraj: Again,

- https://github.com/DragonJAR/Security-Wordlist/blob/main/LFI-WordList-Linux is just a 4 years old copy of the more up to date https://github.com/danielmiessler/SecLists/blob/master/Fuzzing/LFI/LFI-gracefulsecurity-linux.txt
- same with https://github.com/DragonJAR/Security-Wordlist/blob/main/LFI-WordList-Windows outdated copy of https://github.com/danielmiessler/SecLists/blob/master/Fuzzing/LFI/LFI-gracefulsecurity-windows.txt

As for the few files that may not be copied from another source, it would be better to submit them to seclist directly, e.g. https://github.com/DragonJAR/Security-Wordlist/blob/main/TOP-100000-Password-5-Digitos.txt would complete https://github.com/danielmiessler/SecLists/blob/master/Fuzzing/5-digits-00000-99999.txt (DragonJAR list is a frenquency sorted version of the list) or https://github.com/danielmiessler/SecLists/blob/master/Passwords/Common-Credentials/four-digit-pin-codes-sorted-by-frequency-withcount.csv?plain=1 (same but for 4 digits)

PR #4654, opened 2025-10-14
perl-math-pari: fixed error build

D3vil0p3r:
anal patched lol (for readers, it is not a joke)

image


noptrix:
thanks a bunch @D3vil0p3r

Issue #4537, opened 2025-06-02
[Tool request] garak


vdun:
- **Name**: garak
- **Source URL**: https://github.com/NVIDIA/garak
- **Description**: LLM vulnerability scanner

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4657

PR #4657, opened 2025-10-17
garak: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4537


packages/garak/PKGBUILD
```diff
@@ -0,0 +1,50 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=garak
+pkgver=v0.13.1.r11.g8b3a9af
```
noraj: Same as in a previous review about `v`

D3vil0p3r: Fixed

noraj (changes requested):
LGTM, partially tested (I have a too low bandwidth and it take me too much time to re-pull the dependencies at each build)

what a shit ton of dependencies including heavy ones (torch 900 MB)

D3vil0p3r:
I did some basic tests and it seems to work, basically

Issue #4592, opened 2025-08-11
de4dotEx


vdun:
- **Name**: de4dotEx
- **Source URL**: https://github.com/GDATAAdvancedAnalytics/de4dotEx
- **Description**: .NET deobfuscator and unpacker.

noraj:
Same as https://github.com/BlackArch/blackarch/issues/3915:

That's C#, it won't build on Linux nor is something to deploy on a target machine (eg WinPeas).

If you have a proof of it building with [.NET Core](https://wiki.archlinux.org/title/.NET) and [mono-msbuild](https://archlinux.org/packages/?name=mono-msbuild).

vdun:
Can be added as `de4dotEx-bin`

noraj:
Sorry I didn't check, it seems there is a Linux build.

https://github.com/GDATAAdvancedAnalytics/de4dotEx/blob/6023387181702db9091a9742b9e86379cd53c31d/.github/workflows/build.yml#L43

D3vil0p3r:
I was trying to apply the same approach used for [de4dot PKGBUILD](https://github.com/BlackArch/blackarch/blob/master/packages/de4dot/PKGBUILD) but now for it and for de4dotEx, during the build, I get:
```
Determining projects to restore...
/usr/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [/tmp/makepkg/de4dot/src/de4dot/de4dot.netframework.sln]
/usr/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error :   The SSL connection could not be established, see inner exception. [/tmp/makepkg/de4dot/src/de4dot/de4dot.netframework.sln]
/usr/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error :   Authentication failed, see inner exception. [/tmp/makepkg/de4dot/src/de4dot/de4dot.netframework.sln]
/usr/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error :   Ssl error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED [/tmp/makepkg/de4dot/src/de4dot/de4dot.netframework.sln]
[...]

PR #4662, opened 2025-10-18
de4dotex: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4592

### DIFF
--- a/packages/de4dotex/PKGBUILD
+++ b/packages/de4dotex/PKGBUILD
@@ -53,6 +53,6 @@ cd /usr/share/windows/$pkgname
 mono ./de4dot-x64.exe "\$@"
 EOF
 
-  chmod a+x "$pkgdir/usr/bin/$pkgname{,-x64}"
+  chmod a+x "$pkgdir/usr/bin/$pkgname"{,-x64}
 }
 


packages/de4dotex/PKGBUILD
```diff
@@ -0,0 +1,58 @@
...
+    -iname 'Test.Rename.*' -delete
+}
+
+package() {
+  install -dm 755 "$pkgdir/usr/bin"
+  install -dm 755 "$pkgdir/usr/share/$pkgname"
+  install -dm 755 "$pkgdir/usr/share/windows/$pkgname"
+
+  cp -a "$pkgname/Release/net48"/* "$pkgdir/usr/share/$pkgname"
+  install -Dm 755 "$pkgname.sh" "$pkgdir/usr/share/$pkgname/$pkgname.sh"
```
noraj: Instead of creating the 2 scripts in the source tree, you could create them inline in the PKGBUILD like we do for binary wrappers in the jointscript PKGBUILD template: https://github.com/BlackArch/blackarch-pkgbuilds/blob/697fe4d1d7aa2bcf4784f872c5a1eed09b67d5d5/PKGBUILD-generic-jointscript#L47-L51

noraj (changes requested):

packages/de4dotex/PKGBUILD
D3vil0p3r: Fixed

packages/de4dotex/PKGBUILD
```diff
@@ -41,18 +37,24 @@ build() {
...
+  chmod a+x "$pkgdir/usr/bin/$pkgname"
 
-  rm "$pkgdir/usr/share/windows/$pkgname/"{de4dotex.sh,de4dotex-x64.sh}
+  cat > "$pkgdir/usr/bin/$pkgname-x64" << EOF
+#!/bin/sh
+cd /usr/share/windows/$pkgname
+mono ./de4dot-x64.exe "\$@"
+EOF
+
+  chmod a+x "$pkgdir/usr/bin/$pkgname-x64"
```
noraj: Style details, I would done `chmod a+x "$pkgdir/usr/bin/$pkgname{,-x64}"` for both at the same time rather than one command for each.

D3vil0p3r: Yes, it is a cleaner approach.

D3vil0p3r:
> LGTM but please don't trust me and test every time with ba-dev
> 
x) I was by my phone. I'll check by `ba-dev`. EDIT: fixed.

noraj:
I'm sorry I mislead you xD

Issue #4565, opened 2025-07-17
LimeSuite


utenus:
- **Name**: LimeSuite
- **Source URL**: https://github.com/myriadrf/LimeSuite
- - **Description**:  Driver and GUI for LMS7002M-based SDR platforms

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4663

PR #4663, opened 2025-10-18
limesuite: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4565

### DIFF
--- a/packages/limesuite/PKGBUILD
+++ b/packages/limesuite/PKGBUILD
@@ -9,7 +9,7 @@ arch=('x86_64' 'aarch64')
 groups=('blackarch' 'blackarch-radio')
 url='https://github.com/myriadrf/LimeSuite'
 license=('Apache-2.0')
-depends=('glu' 'i2c-tools' 'libusb' 'mesa' 'soapysdr' 'sqlite'
+depends=('glu' 'i2c-tools' 'libusb' 'soapysdr' 'sqlite'
          'wxwidgets-gtk3')
 makedepends=('git' 'cmake')
 source=("git+https://github.com/myriadrf/$pkgname.git")


packages/limesuite/PKGBUILD
```diff
@@ -0,0 +1,45 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=limesuite
+pkgver=v23.11.0.r19.g524cd2e
```
noraj: you can edit the pkgver() like I did in https://github.com/BlackArch/blackarch/pull/4629/files#diff-1231f889678e893a8f0d23d15e31e3322f538cea94fe05b703b041ef817c1b9fR25 to remove the `v` prefix for projects that wrongly add `v` before the version in their git tags.

D3vil0p3r: Fixed

packages/limesuite/PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+
+pkgname=limesuite
+pkgver=23.11.0.r19.g524cd2e
+pkgrel=1
+pkgdesc='Driver and GUI for LMS7002M-based SDR platforms.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-radio')
+url='https://github.com/myriadrf/LimeSuite'
+license=('Apache-2.0')
+depends=('freeglut' 'i2c-tools' 'libusb' 'soapysdr' 'sqlite' 'wxwidgets-gtk3')
```
noraj: Building with ba-dev fails

Image

It seems OpenGL is missing, so `mesa` should be added as a dependency. See https://wiki.archlinux.org/title/OpenGL.

D3vil0p3r: Yeah sorry. The automated workflow was based on the wrong image. Now it is fixed and the next PRs should not have this inconsistency as the previous commit in this PR.

noraj: Sorry, I was wrong, mesa was not needed, it was `glu` that was needed in addition to `freeglut`. Else `usr/include/GL/glu.h` is missing and `OpenGL::GLU` is not satisfied. Please also test if `freeglut` is needed at all.

D3vil0p3r: I tested by `ba-dev` my first proposal without `mesa` and it seems to work. I don't know why it behaves in this manner. Btw I tested it in a containerized minimum env and yes, `glu` is missing.

noraj: I see you replaced freeglut with glu, ok if it works, but I thing yo ucan remvoe mesa too.

I tested with the following (no mesa) an it worked.

```
depends=('freeglut' 'i2c-tools' 'libusb' 'glu' 'soapysdr' 'sqlite'
         'wxwidgets-gtk3')
```

D3vil0p3r: Yeah, I tested on container without `mesa`, I forgot to remove it from the PKGBUILD here. Now it is fixed.

Issue #4564, opened 2025-07-17
deskhpsdr


utenus:
- **Name**: deskhpsdr
- **Source URL**: https://github.com/dl1bz/deskhpsdr
- **Description**:  SDR App for HPSDR (old and new) protocol and Soapy-API (macOS and Linux)

D3vil0p3r:
https://github.com/BlackArch/blackarch/pull/4664

PR #4664, opened 2025-10-18
deskhpsdr: add package

D3vil0p3r:
Close https://github.com/BlackArch/blackarch/issues/4564


D3vil0p3r:
Because its [Makefile](https://github.com/dl1bz/deskhpsdr/blob/master/Makefile) is poorly designed. Btw I patch the Makefile to not allow the creation of the icon system dir.

noraj (approved):
Nice, should be perfect now.

Issue #4453, opened 2025-01-25
caido-desktop


sys-faz:
- **Name**: caido-desktop
- **Source URL**: https://caido.io/
- **Description**: A lightweight web security auditing toolkit.

noraj:
We avoid distributing pre-compiled binaries and since it's a closed-source software, they are no source we can build.

Also, it is unclear if we can legally redistribute this proprietary software as there is no [OSI license](https://opensource.org/licenses) and the legal terms are the following.

> You may not reproduce or redistribute any software except in accordance with the EULA or these Terms and Conditions.
>
> https://caido.io/fr/terms

sys-faz:
got it.

PR #4666, opened 2025-10-19
caido: add package

D3vil0p3r:


noraj:
See https://github.com/BlackArch/blackarch/issues/4453 and https://github.com/BlackArch/blackarch/issues/4454.

The issue was we were not allowed to redistribute the binaries. 

However, I discussed with Emile and Christopher while they were doing their Caido workshop at HackFest, and they'll allow us to redistribute it in BlackArch. 

Let me check before merging this. They told me they have an API listing new versions so we maybe able to use it in a pkgver().

packages/caido/PKGBUILD
```diff
@@ -0,0 +1,50 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=caido
+pkgname=('caido-cli' 'caido-desktop')
+pkgver=0.52.0
+pkgrel=1
+pkgdesc='A lightweight web security auditing toolkit.'
```
noraj:
```suggestion
pkgdesc='Intercepting proxy to replay, inject, scan and fuzz HTTP requests.'
```

We are not forced to copy upstream description, I find something like `Intercepting proxy to replay, inject, scan and fuzz HTTP requests` clearer than a generic `web security auditing toolkit`.

packages/caido/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+url='https://caido.io/'
+license=('custom:unknown')
+arch=('x86_64' 'aarch64')
+source_x86_64=(
```
noraj: I never tested, would that be possible to put each `source` and `sha512sums` inside `package_caido-desktop` and `package_caido-cli` ?

packages/caido/PKGBUILD
```diff
@@ -0,0 +1,50 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgbase=caido
+pkgname=('caido-cli' 'caido-desktop')
+pkgver=0.52.0
+pkgrel=1
+pkgdesc='A lightweight web security auditing toolkit.'
+url='https://caido.io/'
+license=('custom:unknown')
```
noraj:
```suggestion
license=('custom:private')
```

It would be also better to scrap https://caido.io/fr/terms and copy it to `install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"`

packages/caido/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+sha512sums_x86_64=('88ed092074b0a933744412c187bed07e76ca3c4a6a8e327302a732baf8a87d5af8fa7e7a1e7e9f09459ce934fde455002b19419b9a2dc2aed9dc37ee5e8b4713'
+                   '6dd7c244d40658c3fa0ff89a89f36fb3ede3db4ead9e8a24e5067b9c90835c923a255ad05836a4feace90a3e18735efa743faa68f9398b30e8c093d99e937b9e')
+sha512sums_aarch64=('38bd8938feef678fe18965558bbdc859331e2a77a3df03b167f6467313d04717e18de7caaa2752c3a117342cdc0dc5144e0f97f1931e8ff75ee6aaab1bb2bfb9'
+                    '7a1780e8acbe2848daa7c60ac4779f9547862ddfc9aa66ad649e3aafcf05c4911b72142284574214277f5a92cc22c529b266d7244d50c63fd7cd1b74b7569581')
+
+package_caido-desktop() {
+  depends=('glibc' 'gcc-libs')
+  provides=('caido')
+
+  chmod +x *.AppImage
```
noraj: Follow BlackArch template with a `prepare_caido-desktop()` and `build_caid-desktop()` rather than just copying the AUR.

https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-appimage

packages/caido/PKGBUILD
D3vil0p3r: Agreed

packages/caido/PKGBUILD
D3vil0p3r: Sadly no:
```
==> ERROR: source can not be set inside a package function
==> ERROR: sha512sums can not be set inside a package function
==> ERROR: source can not be set inside a package function
==> ERROR: sha512sums can not be set inside a package function
```
Btw I simplified the source and sha512sums arch management.

packages/caido/PKGBUILD
D3vil0p3r: Done. For the LICENSE I asked Emile in Caido Discord server. Meanwhile I uploaded a `LICENSE` file from source with the content from https://caido.io/terms. Let me know it is ok.

packages/caido/PKGBUILD
noraj: Hard copy of the Terms is a nice workaround, but it means it won't be updated if it changes. Let's wait a bit to see if they can upload a file somewhere that we can download.

noraj: Based on the recent discussion on Discord, let's go with this manual copy approach since it will take time before something changes.

noraj (approved):
LGTM

PR #4709, opened 2025-11-27
semgrep: fix venv upgrade

noraj:
see https://github.com/BlackArch/blackarch/pull/4709 more generally

I noticed I had semgrep package version 1.139.0 installed but it displayed `1.101.0` when doing `semgrep --version` and indeed when running `post_install()` again manually, it displayed that dependencies are already satisfied as 1.101.0 was already installed. So all pip venv package are actually never updating and stay in the version they were first installed in.

We should update the PKGBUILD of all python venv packages.

### DIFF
--- a/packages/semgrep/PKGBUILD
+++ b/packages/semgrep/PKGBUILD
@@ -2,7 +2,7 @@
 # See COPYING for license details.
 
 pkgname=semgrep
-pkgver=1.139.0
+pkgver=1.144.0
 pkgrel=1
 epoch=1
 pkgdesc='Lightweight static analysis for many languages.'


noptrix:
> see #4709 more generally
> 
> I noticed I had semgrep package version 1.139.0 installed but it displayed `1.101.0` when doing `semgrep --version` and indeed when running `post_install()` again manually, it displayed that dependencies are already satisfied as 1.101.0 was already installed. So all pip venv package are actually never updating and stay in the version they were first installed in.
> 
> We should update the PKGBUILD of all python venv packages.

good catch. thanks

PR #4723, opened 2025-12-04
nuclei: remove forced -config flag.

s7x:
A small fix to prevent nuclei using the default config file in /usr/share/nuclei/config.yaml.
Cheers!

### DIFF
--- a/packages/nuclei/PKGBUILD
+++ b/packages/nuclei/PKGBUILD
@@ -53,7 +53,7 @@ package() {
 
   cat > "$pkgdir/usr/bin/$pkgname" << EOF
 #!/bin/sh
-exec /usr/share/$pkgname/$pkgname -config /usr/share/$pkgname/config.yaml "\$@"
+exec /usr/share/$pkgname/$pkgname "\$@"
 EOF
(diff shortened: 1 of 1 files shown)


packages/nuclei/PKGBUILD
```diff
@@ -53,7 +53,7 @@ package() {
 
   cat > "$pkgdir/usr/bin/$pkgname" << EOF
 #!/bin/sh
-exec /usr/share/$pkgname/$pkgname -config /usr/share/$pkgname/config.yaml "\$@"
+exec /usr/share/$pkgname/$pkgname "\$@"
```
noraj, line 56: Please revert. It was intended, it's is not a default configuration but a [custom configuration](https://github.com/BlackArch/blackarch/blob/master/packages/nuclei/config.yaml) that is deployed. It was added in https://github.com/BlackArch/blackarch/pull/4357 to force nuclei to use nuclei-template package and resolve this issue https://github.com/BlackArch/blackarch/issues/3382. Else, nuclei deploys ugly folder at the root of the user home directory.

PR #4726, opened 2025-12-04
nuclei: revert s7x change

noraj:
Revert https://github.com/BlackArch/blackarch/pull/4723 as explained in https://github.com/BlackArch/blackarch/pull/4723#discussion_r2590647285

### DIFF
--- a/packages/nuclei/PKGBUILD
+++ b/packages/nuclei/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=nuclei
 pkgver=v3.4.7.r281.gf181a69
-pkgrel=1
+pkgrel=2
 epoch=2
 groups=('blackarch' 'blackarch-scanner' 'blackarch-webapp' 'blackarch-recon'
         'blackarch-fingerprint')
@@ -53,7 +53,7 @@ package() {
 
   cat > "$pkgdir/usr/bin/$pkgname" << EOF
 #!/bin/sh
-exec /usr/share/$pkgname/$pkgname "\$@"
+exec /usr/share/$pkgname/$pkgname -config /usr/share/$pkgname/config.yaml "\$@"
 EOF
 
   chmod +x "$pkgdir/usr/bin/$pkgname"


s7x:
We should add a disclaimer then if we revert this, just in case people start wondering why is nuclei not taking into account their config file located in ~/.config/nuclei/config.yaml. What do you think @noraj ?

noraj:
@s7x The issue is nuclei only follows [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) ([[AL](https://wiki.archlinux.org/title/XDG_Base_Directory)]) partially. It only looks at `~/.config/nuclei/config.yaml` and does not resolve `$XDG_CONFIG_DIRS` or `$XDG_CONFIG_HOME`, that would allow us to deploy the config at `/etc/nuclei/config.yml` globally for all user and then each user to override with their own `~/.config/nuclei/config.yaml` if wanted.

Because XDG standard is not followed for data dir, we made a config wrapper that is now hurting the wall of the config dir because XDG standard is not followed here either. We could continue to add patches and wrappers, but it would be best that upstream follows XDG standard instead.

I have opened a discussion upstream: https://github.com/orgs/projectdiscovery/discussions/6658

s7x:
> @s7x The issue is nuclei only follows [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/) ([[AL](https://wiki.archlinux.org/title/XDG_Base_Directory)]) partially. It only looks at `~/.config/nuclei/config.yaml` and does not resolve `$XDG_CONFIG_DIRS` or `$XDG_CONFIG_HOME`, that would allow us to deploy the config at `/etc/nuclei/config.yml` globally for all user and then each user to override with their own `~/.config/nuclei/config.yaml` if wanted.
> 
> Because XDG standard is not followed for data dir, we made a config wrapper that is now hurting the wall of the config dir because XDG standard is not followed here either. We could continue to add patches and wrappers, but it would be best that upstream follows XDG standard instead.
> 
> I have opened a discussion upstream: https://github.com/orgs/projectdiscovery/discussions/6658

I did not know they were not following the XDG entirely. Thanks for the insight & the PR/Issue!

noraj:
They added XDG support https://github.com/projectdiscovery/nuclei/pull/7684, we should look at what we can do now

PR #4733, opened 2025-12-11
pkgcheck: fixed error misuse if $pkgname is with no double quotes

D3vil0p3r:
Currently all variables need to have double quotes otherwise pkgcheck triggers an error. This is wrong. Only $pkgdir and $srcdir need double quotes. This PR fixes this issue.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,1 +1,0 @@
-pkgcheck-arch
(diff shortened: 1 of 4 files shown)


noraj (approved):
LGTM

PR #4804, opened 2026-01-02
Add evil-winrm-py

0xbbuddha:
This PR adds **evil-winrm-py**, a Python implementation of a WinRM shell for Windows and Active Directory pentesting.

- Upstream: https://github.com/adityatelange/evil-winrm-py
- Installs the following binaries:
  - evil-winrm-py
  - ewp
- Grouped under `blackarch` and `blackarch-windows`
- Built and tested locally with `makepkg`

Let me know if any adjustments are needed.

### DIFF
--- a/packages/evil-winrm-py/PKGBUILD
+++ b/packages/evil-winrm-py/PKGBUILD
@@ -28,3 +28,4 @@ package() {
     install -Dm644 LICENSE "${pkgdir}/usr/share/licenses/${pkgname}/LICENSE"
   fi
 }
+


noraj:
The tool doesn't seem to bring much features compared to the original https://github.com/Hackplayers/evil-winrm. Upstream (rb evil-winrm) is still maintained, this project (py evil-winrm) is just a python port. It's (py) not easier to use or more accessible that way. However, it's true the original tool (rb) has a very bad architecture (2k LoC long one-file) which make it hard to contribute to it but easy to drop on a remote system. But anyway, that tool is still featureful and easy to use, I don't see interest into packaging a redundant tool doing exactly the same for end users.

0xbbuddha:
Thanks for the detailed feedback, that makes sense.

I agree that from an end-user perspective the Ruby version already covers the use case very well and is still actively maintained, and I understand the concern about avoiding redundant tools in BlackArch.

One motivation for this PR was that Ruby can sometimes be an obstacle for some users or environments, whereas Python can make the tool a bit more friendly and accessible in certain setups. That said, I agree this alone may not be a strong enough reason to justify packaging a redundant tool.

No problem closing this PR if you feel it doesn’t fit the project’s scope.
Thanks for taking the time to review it.

noraj:
> One motivation for this PR was that Ruby can sometimes be an obstacle for some users or environments, whereas Python can make the tool a bit more friendly and accessible in certain setups. That said, I agree this alone may not be a strong enough reason to justify packaging a redundant tool.

I may be true outside distro packaging systems, for people wanting to install tools manually, but that precisely one of the interests of distro packaging, either way, whatever the programming language is behind, the BlackArch user will install it with `pacman -S` in both cases and don't have to mind how the language ecosystem is working.

> No problem closing this PR if you feel it doesn’t fit the project’s scope.

I'll let the others decide. But thank you for your contribution anyway.

ikstream:
I had some engagements recently where evil-winrm-py was the tool of choice, as evil-winrm behaved unstable and connections have been lost. 
So it might be a good alternative.

unfortunately we hadn’t had enough time to investigate what caused the instability.

D3vil0p3r:
@0xbbuddha meanwhile please check the `pkgcheck` workflow error and fix it.

0xbbuddha:
> @0xbbuddha meanwhile please check the `pkgcheck` workflow error and fix it.

done

0xbbuddha:
I investigated the runtime error and found that python-pypsrp is currently built for Python 3.13.

The package installs its modules in:
`/usr/lib/python3.13/site-packages/pypsrp`
instead of `/usr/lib/python3.14/site-packages/pypsrp`.

As a result, `import pypsrp` fails even though python-pypsrp is installed.

Do you know if python-pypsrp will be rebuilt for Python 3.14 soon?

D3vil0p3r:
In theory it's needed only to rebuild the package on a server including Python 3.14, and hoping that the library is compatible with latest version. Btw currently a workaround for you would be just copy that library to the 3.14 directory that you can manually create.

0xbbuddha:
Yes, in the meantime, the solution is to copy the pypsrp module from Python 3.13 to Python 3.14:
```bash
sudo cp -r /usr/lib/python3.13/site-packages/pypsrp /usr/lib/python3.14/site-packages/
```

PR #4806, opened 2026-01-03
Python packages migrated to python-installer from python-pip.

KevinCrrl:
In accordance with discussion #4800, I propose these 5 packages as a sample of a series of successful tests to use python-installer instead of pip in accordance with modern standards.

General changes:

- makedepends python-pip replaced by python-installer and python-setuptools
- pkgrel++

Other changes:
 - Fix typo
 - Fix missing depend in pyinstaller


D3vil0p3r:
Can you please check the output of the failing workflow `pkgcheck` and fix the mentioned errors?

PR #4813, opened 2026-01-14
New package: pypy3-flit-core

chouettevan:
This PR, which should be followed by a series of other PRs, has the goal of adding a pypy-angr package which allows for running angr under pypy. according to the angr docs,using pypy allows for 10x performance improvement (see the [docs](https://docs.angr.io/en/latest/advanced-topics/speed.html) ) .pypy3-flit-core will allow the addition of the pypy build system, which should be followed with angr's dependencies, and finally angr itself

### DIFF
--- a/packages/pypy3-flit-core/PKGBUILD
+++ b/packages/pypy3-flit-core/PKGBUILD
@@ -1,0 +1,23 @@
+#  
+_base=flit_core
+pkgname=pypy3-${_base//_/-}
+pkgver=3.12.0
+pkgrel=1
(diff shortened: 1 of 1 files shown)


noptrix:
@chouettevan thanks! please check my last changes on this for our style guidelines when submitting future PRs: https://github.com/BlackArch/blackarch/commit/747016cc58c242a01ada53f50b919b472f780026

there are also PKGBUILD templates available in the `blackarch-pkgbuilds` repository that you can refer to.

PR #4815, opened 2026-01-15
Angr fix

chouettevan:
After the update to angr, it turns out that newer versions of angr have new  runtime dependencies. This PR adds the newly needed dependencies.


D3vil0p3r:
Remember to check the `pkgcheck` workflow errors.

PR #4817, opened 2026-01-15
Python-arpspoof: Fix typo and migrating from pip to python-installer

KevinCrrl:
#4800


noptrix:
@KevinCrrl thanks a lot

PR #4822, opened 2026-01-16
hexodus: add package and remove breads

D3vil0p3r:
`breads` has been revamped as `hexodus`.


packages/hexodus/PKGBUILD
```diff
@@ -0,0 +1,51 @@
...
+build() {
+  cd $pkgname
+
+  python -m build --wheel --outdir="$startdir/dist"
+}
+
+package() {
+  cd $pkgname
+
+  pip install \
```
noraj: Do we try python-installer?

https://github.com/BlackArch/blackarch/discussions/4800

D3vil0p3r: Do you mean by using:
```
build() {
  cd "$pkgname-$pkgver"

  python -m build --wheel --no-isolation
}

package() {
  cd "$pkgname-$pkgver"

  python -m installer --destdir="$pkgdir" dist/*.whl
}
```
?

PR #4828, opened 2026-01-17
Migrating packages from pip to python-installer

KevinCrrl:
Migrated packages in to-release; #4800


noptrix:
@KevinCrrl thanks a lot

PR #4831, opened 2026-01-20
Added TWDS Mirror

seadog007:
The mirror is there since 2024/08, forgot to send the PR.


noraj (approved):

D3vil0p3r:
You should edit as well also here: https://github.com/BlackArch/blackarch-site/blob/master/blackarch-mirrorlist

Issue #4816, opened 2026-01-15
Package request: glpwnme


noraj:
### Describe the package

N/A

### Source URL

https://github.com/Orange-Cyberdefense/glpwnme

### License

N/A

### Have you considered submitting a pull request?

y

### I confirm that this request is relevant for BlackArch

- [x] I assert that this package request does not yet exist in an [open pull request](https://github.com/BlackArch/blackarch/pulls?q=is%3Aopen).
- [x] I assert that this is not a [duplicate of an existing request](https://github.com/BlackArch/blackarch/issues?q=is%3Aopen%20is%3Aissue%20label%3Atype%3A%3Atool-request). 
- [x] I assert that the package is not available in BlackArch or official [Arch Linux repositories](https://archlinux.org/packages/).
- [x] I have reviewed the [PKGBUILD templates](https://github.com/BlackArch/blackarch-pkgbuilds) and understand that pull requests are preferred when possible.
- [x] I assert that I have read the [Arch Linux Code of Conduct](https://terms.archlinux.org/docs/code-of-conduct/) and agree to abide by it.

KevinCrrl:
#4834

PR #4834, opened 2026-01-20
glpwnme: adding package

KevinCrrl:
#4816 

I created this PKGBUILD for the glpwnme package, but I'm not sure whether to leave it as it is. I welcome any suggestions or corrections :)


noraj (commented):
Thanks.

PR #4838, opened 2026-01-28
onionsearch: fix package.

s7x:
Fixed the onionsearch package that wasn't working and was missing deps.

I used a mix of the PKGBUILD-python-standalone and the PKGBUILD-python-PEP517-setuptools templates:

- It fetches the version from the git
- Installs the README at the proper location
- Uses the python libs from the repo so it's not cluttering with a venv

Cheers!

EDIT: typo

### DIFF
--- a/packages/onionsearch/PKGBUILD
+++ b/packages/onionsearch/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=onionsearch
 pkgver=44.fc9d62c
-pkgrel=1
+pkgrel=5
 pkgdesc='Script that scrapes urls on different .onion search engines.'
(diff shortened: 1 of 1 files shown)


packages/onionsearch/PKGBUILD
```diff
@@ -3,34 +3,41 @@
 
 pkgname=onionsearch
 pkgver=44.fc9d62c
-pkgrel=4
-pkgdesc='Script that scrapes urls on different ".onion" search engines.'
+pkgrel=1
```
noraj:
```suggestion
pkgrel=5
```

pkgrel++ if pkgver doesn't move

PR #4925, opened 2026-04-27
Update python packages

KevinCrrl:
Simple updates

Only the "export CFLAGS" line is removed from the pyinstaller PKGBUILD, as it was a fix applied in #4857 and has already been resolved in https://github.com/pyinstaller/pyinstaller/pull/9372


noptrix:
thank you @KevinCrrl

PR #4926, opened 2026-04-28
Update Calamares to version 3.4.2.26.0427

Zen1th53:
### fix: calamares build (qt6 / kpmcore)

calamares was failing to build on current arch due to a qt6/kpmcore mismatch and test linking errors. this pr fixes that.

**changes:**

* disable tests (causing kpmcore/qt6 symbol issues)
* switch to ninja (`cmake --build`)
* clean up pkgbuild (drop unneeded alci-specific bits)
* use proper install via `cmake --install`

**result:**

* builds clean on up-to-date arch
* installs without conflicts
* runs as expected (with external config)

---

this also helps restore the calamares installer for the blackarch slim iso.

---

would appreciate a review and merge. tested locally.
image
calamare


noraj:
Thank you, looks promising. I'll let someone more knowledgeable in C builds review that PR.

nextboxis (approved):

PR #4253, opened 2024-10-07
faradaysec: added missing dependencies

D3vil0p3r:
https://github.com/BlackArch/blackarch/issues/3660

PR #4927, opened 2026-04-28
Python selenium update

KevinCrrl:
Update Python-Selenium: major version update 3.x -> 4.x
- Migrating package: python-installer instead python-pip
- Fix build and depends: Using rust depends for new modules
- Fix arch for compiled packages

Fix attacksurfacemapper
- Using new pkgver function
- Patching to use python-selenium 4.x
- adding epoch

Fix httpscreenshot
- Patching to use selenium 4

Migrating python-splinter
- fix build for python3-only setup.py

Faradaysec update reverted in favor of #4253 

Fix traxss:
- Fix license
- Update pkgver function
- patch for selenium 4.x

I tested a few other packages but it seems they were already adapted to Selenium 4, even so if in the future I find a bug in a Selenium-related package, I will try to fix it.

I assume the PR is ready in these packages, but I'm always willing to correct anything that's wrong.


noptrix:
@KevinCrrl thank you very much

PR #4964, opened 2026-06-13
fix full repository installation conflicts

### LINKED ISSUE #4960: Bug: Unable to install entire repository
### BlackArch version

- BlackArch Linux 64 bit Full ISO 2023.04.01

### Which environment or technology are you using?

- [x] Baremetal
- [ ] VirtualBox Virtual Machine
- [ ] VMware Virtual Machine
- [ ] Hyper-V Virtual Machine
- [ ] QEMU/KVM
- [ ] Container

### Kernel version

_No response_

### Describe the bug

Hey guys - I'm installing blackarch over an existing arch installation that I dropped in place a few days ago. In the past, it's gone pretty seamlessly, but I'm having some issues installing the entire blackarch repo as I've done in the past.

The first was related to libsoup3. I had some dependencies on the previous version. Fixed by installing libsoup via AUR.

I have tons of warnings about an unknown key 'makepkgopt' in package descriptions, but this is just annoying, not a blocker.

The second resulted from multiple tools using /usr/lib/node_modules/*, which I solved by adding the --overwrite flag (i.e., --overwrite "/usr/lib/node_modules/*"). I'm now stuck with a conflict between metasploit and viproy-voipkit, which is not solved by adding a an "--overwrite /usr/share/doc/metaspoit/*" tag. I can't seem to get around this one easily, so I'm changing my tool installation strategy from everything to just particular categories of tools.

I just wanted to let you know what I had encountered - it seems no one had mentioned any of this yet.

BTW, I'd also suggest that the strap.sh script also install the multilib dependencies.

### Steps to reproduce

1. Install latest Arch Linux.
2. install blackarch repo via strap.sh.
3. Execute "sudo pacman -Syu", select default options for prompts and y to questions.

The install should fail.

### Expected behaviour

The install of blackarch tools succeeds with no errors or warnings.

### Screenshots

_No response_

### Relevant log output

```console

```

### I confirm that this issue is relevant for BlackArch

- [x] I assert that this is a bug and not a support request.
- [x] I assert that this is not a [duplicate of an existing issue](https://github.com/BlackArch/blackarch/issues?q=is%3Aopen%20is%3Aissue%20label%3Atype%3A%3Abug). 
- [x] I assert that I have read the [Arch Linux Code of Conduct](https://terms.archlinux.org/docs/code-of-conduct/) and agree to abide by it.

remxs2k:
## Summary

- Add the libsoup 2 compatibility package required by `webkit2gtk`.
- Remove obsolete `npm --user root` arguments that created conflicting `root` modules.
- Stop `viproy-voipkit` from owning paths belonging to `metasploit`.
- Bump affected package release numbers.

## Testing

- Validated all changed PKGBUILDs with `bash -n` and `makepkg --printsrcinfo`.
- Successfully built and inspected `padoracle` and `viproy-voipkit`.
- Verified the libsoup sources and checksums with `makepkg --verifysource`.

Fixes #4960

### DIFF
--- a/packages/git-dump/PKGBUILD
+++ b/packages/git-dump/PKGBUILD
@@ -4,7 +4,7 @@
 pkgname=git-dump
 _pkgname=node-git-dump
 pkgver=7.4c9a2a9
-pkgrel=1
+pkgrel=2
 pkgdesc='Dump the contents of a remote git repository without directory listing enabled.'
 arch=('any')
 groups=('blackarch' 'blackarch-scanner' 'blackarch-code-audit')
@@ -30,7 +30,7 @@ package() {
 
   cd "$_npmdir"
 
-  npm install --user root -g --prefix "$pkgdir/usr" "$pkgname@"
+  npm install -g --prefix "$pkgdir/usr" "$pkgname@"
 
   cd "$srcdir/$_pkgname"
 
--- a/packages/bagbak/PKGBUILD
+++ b/packages/bagbak/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=bagbak
 pkgver=342.71fbd4d
-pkgrel=1
+pkgrel=2
 pkgdesc='Yet another frida based App decryptor.'
 arch=('any')
 groups=('blackarch' 'blackarch-mobile' 'blackarch-reversing' 'blackarch-binary')
@@ -29,7 +29,7 @@ package() {
 
   cd "$_npmdir"
 
-  npm install --user root -g --prefix "$pkgdir/usr" "$pkgname@"
+  npm install -g --prefix "$pkgdir/usr" "$pkgname@"
 
   cd "$srcdir/$pkgname"
 
--- a/packages/brosec/PKGBUILD
+++ b/packages/brosec/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=brosec
 pkgver=278.c51164f
-pkgrel=1
+pkgrel=2
 groups=('blackarch' 'blackarch-exploitation')
 pkgdesc='An interactive reference tool to help security professionals utilize useful payloads and commands.'
 url='https://github.com/gabemarshall/Brosec'
@@ -30,7 +30,7 @@ package(){
 
   cd "$_npmdir"
 
-  npm install --user root -g --prefix "$pkgdir/usr" Brosec@
+  npm install -g --prefix "$pkgdir/usr" Brosec@
 
   cd "$srcdir/$pkgname"
 
--- a/packages/express/PKGBUILD
+++ b/packages/express/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=express
 pkgver=4.17.3
-pkgrel=1
+pkgrel=2
 epoch=1
 pkgdesc='High performance, high class web development for node.js.'
 arch=('any')
@@ -23,7 +23,7 @@ package() {
 
   cd $_npmdir
 
-  npm install --user root -g --prefix "$pkgdir/usr" "express@$pkgver"
+  npm install -g --prefix "$pkgdir/usr" "express@$pkgver"
   rm -rf "$_npmdir/root"
 }
 
--- a/packages/jstillery/PKGBUILD
+++ b/packages/jstillery/PKGBUILD
@@ -3,7 +3,7 @@
 
 pkgname=jstillery
 pkgver=65.512e9af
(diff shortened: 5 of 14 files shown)


noraj (commented):
Thanks, this is interesting. Libsoup will probably need some adaptation from the AUR.

remxs2k:
> Thanks, this is interesting. Libsoup will probably need some adaptation from the AUR.

Thanks. The libsoup PKGBUILD in this PR is based on the current AUR libsoup 2.74.3-4 recipe. I kept the AUR source tag, dependency/provides set, docs split, and CVE cherry-pick, and adapted it for BlackArch style/aarch64. The flaky test disable is applied inline in prepare() instead of carrying the AUR patch file.

Happy to adjust it closer to the AUR version if preferred.

packages/git-dump/PKGBUILD
```diff
@@ -30,11 +30,10 @@ package() {
...
 
-  npm install --user root -g --prefix "$pkgdir/usr" "$pkgname@"
+  npm install -g --prefix "$pkgdir/usr" "$pkgname@"
 
   cd "$srcdir/$_pkgname"
 
   install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
   install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
 }
-
```
noraj, line 40: your editor must be auto-removing the trailing blank like and no following EditorConfig (https://github.com/BlackArch/blackarch/blob/master/.editorconfig), please make your IDE follows EditorConfig

packages/libsoup/PKGBUILD
```diff
@@ -0,0 +1,58 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+#
+# Maintainer: Jan Alexander Steffens (heftig) <heftig@archlinux.org>
+# Contributor: Jan de Groot <jgc@archlinux.org>
+
+pkgbase=libsoup
```
noraj, line 4: Could you explain why would we need to package old libsoup 2.x while AL is already officially packaging [libsoup 3.x](https://archlinux.org/packages/extra/x86_64/libsoup3/)?

noraj (changes requested):

packages/libsoup/PKGBUILD
noraj: Ah yes for https://github.com/BlackArch/blackarch/blob/master/packages/webkit2gtk/PKGBUILD, @noptrix, could you tell why you added webkit2gtk?

noraj: Anyway the AUR package is outdated, webkit2gtk requires libsoup 3 not libsoup 2: https://github.com/search?q=repo%3AWebKit%2FWebKit%20libsoup&type=code

noraj: Also @noptrix can't we just rely on [webkit2gtk-4.1](https://archlinux.org/packages/extra/x86_64/webkit2gtk-4.1/)? https://aur.archlinux.org/pkgbase/webkit2gtk is outdated

remxs2k: Amm... yeah, sorry about the blank lines, I will restore those.
Now, about libsoup, I added libsoup 2 because the current BlackArch `packages/webkit2gtk/PKGBUILD` still builds with `-D USE_SOUP2=ON` and provides the legacy `libwebkit2gtk-4.0.so`, so the existing BlackArch package currently requires libsoup 2 rather than libsoup3.

That said, I agree this is only a compatibility fix. If the preferred direction is to rely on Arch’s `webkit2gtk-4.1`/libsoup3 instead, I can remove the libsoup addition from this PR and keep only the npm/root-module and viproy/metasploit conflict fixes

Fawz-Haaroon: commit 7208fc6b8bfe1ce701444bc1719e5151a563ab4b updates webkit2gtk to use libsoup3 and removes the obsolete USE_SOUP2 option

D3vil0p3r: > Also @noptrix can't we just rely on [webkit2gtk-4.1](https://archlinux.org/packages/extra/x86_64/webkit2gtk-4.1/)? https://aur.archlinux.org/pkgbase/webkit2gtk is outdated

Some tools and libraries work with only webkit2gtk and not with webkitgtk2-4.1. The former is still needed for some old tools.

PR #4966, opened 2026-06-15
android-backup-extractor: initial commit

smiley8D:
## Before submitting, please select the type of PR you are opening

- [X] 📦 New tool/library submission
- [ ] 🕸️ New mirror submission

---

# 📦 New tool/library submission

### Package name
android-backup-extractor

### Description
Short-and-simple cmdline tool to unpack or pack android backups (.ab). It supports passwords for encrypted backups.

### Source URL
https://github.com/nelenkov/android-backup-extractor

### License
Apache 2.0

### Tool category
mobile

### Additional context
I discovered this while solving some HackTheBox challenges. It's very straightforward and does the job well.

### Checklist

- [X] I assert that this tool is not already available in BlackArch or the official [Arch Linux repositories](https://archlinux.org/packages/).
- [X] I assert that this is not a duplicate of an [existing open PR](https://github.com/BlackArch/blackarch/pulls?q=is%3Aopen).
- [X] I assert that I have successfully tested, built and installed the package locally.
- [X] I assert that the PKGBUILD follows the official [BlackArch PKGBUILD templates](https://github.com/BlackArch/blackarch-pkgbuilds).
- [X] I assert that the tool is actively maintained upstream and does not depend on deprecated languages or libraries.
- [X] I assert that I have read the [Arch Linux Code of Conduct](https://terms.archlinux.org/docs/code-of-conduct/) and agree to abide by it.


noraj (changes requested):
Please follow this template https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-java-maven

smiley8D:
Done! This no longer appears to pin the package to the version in PKGBUILD, as it will just build the latest commit. Unless I'm mistaken, I also cannot set a sha512 hash for the source either.

Is this acceptable/common?
Also, is java-maven generally preferred over java if possible? Is it a matter of auditability?

Thx :)

noraj:
> Done! This no longer appears to pin the package to the version in PKGBUILD, as it will just build the latest commit. Unless I'm mistaken, I also cannot set a sha512 hash for the source either.
> 
> Is this acceptable/common?

Yes, there is SKIP for hashing when using git updates as it's rebuild for each commit and not downloading an archive.

> Also, is java-maven generally preferred over java if possible? Is it a matter of auditability?

It's better to build from source rather than downloading a pre-built binary where there is no guarantee it matches the source code.

packages/android-backup-extractor/PKGBUILD
```diff
@@ -0,0 +1,54 @@
...
+package() {
+  cd $pkgname
+
+  install -dm 755 "$pkgdir/usr/bin"
+
+  install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md
+
+  # Only install license if required according to
+  # https://wiki.archlinux.org/index.php/PKGBUILD#license
+  install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
```
noraj: If you check I think Apache-2.0 is in the folder and all this can be removed.

noraj (changes requested):

packages/android-backup-extractor/PKGBUILD
smiley8D: Should I remove just the comments and keep line 43?

noraj: Read the comment, then the wiki and you should understand 😀

smiley8D: Derp, dropped the Apache license install :P

noraj (approved):
LGTM

smiley8D:
> LGTM

What's the right way to bump noptrix? Looks like they need to approve as well?

noraj:
You can try matrix. But I think he's overbusy right now and a bit AFK. I'm sorry you have to wait for merging.

Issue #4974, opened 2026-06-30
Bug: NetExec fails with ModuleNotFoundError: dploot.triage.cng after system update


0xN4no:
### BlackArch version

- BlackArch Linux 64 bit Full ISO 2023.04.01

### Which environment or technology are you using?

- [x] Baremetal
- [ ] VirtualBox Virtual Machine
- [ ] VMware Virtual Machine
- [ ] Hyper-V Virtual Machine
- [ ] QEMU/KVM
- [ ] Container

### Describe the bug

After updating my Arch Linux system (with BlackArch repository enabled), NetExec stopped working.

Command used:
nxc smb 172.16.139.3 --shares

Error:
ModuleNotFoundError: No module named 'dploot.triage.cng'

Environment:
- OS: Arch Linux (BlackArch repository enabled)
- netexec: v1.5.0.r6.g6ae1f0b (BlackArch package)
- python-dploot: 3.2.2-1 (Arch package)

Expected behavior:
NetExec should run normally or handle dependency mismatch gracefully.

Actual behavior:
NetExec crashes during import due to missing module in current dploot package.

Additional notes:
The issue started immediately after a system update (pacman -Syu).

<img width="681" height="231" alt="Image" src="https://github.com/user-attachments/assets/3f7b8571-b25d-4283-9972-a567c83ba964" />

### Steps to reproduce

1. Install NetExec from BlackArch/Arch repositories (sudo pacman -S netexec)
2. Install python-dploot from Arch repositories
3. Update system using pacman -Syu
4. Run NetExec command:
   nxc smb <target> --shares
5. Observe crash with ModuleNotFoundError

### Expected behaviour

NetExec should execute normally without crashing or should detect incompatible dependency versions and fail gracefully with a clear error message.

### Relevant log output

```console
Traceback (most recent call last):
  File "netexec.py", line 231, in <module>
  File "nxc/loaders/protocolloader.py", line 14, in load_protocol
  File "/tmp/_MEI*/nxc/protocols/smb.py", line 82, in <module>
    from dploot.triage.cng import CngTriage
ModuleNotFoundError: No module named 'dploot.triage.cng'
```

### I confirm that this issue is relevant for BlackArch

- [x] I assert that this is a bug and not a support request.
- [x] I assert that this is not a [duplicate of an existing issue](https://github.com/BlackArch/blackarch/issues?q=is%3Aopen%20is%3Aissue%20label%3Atype%3A%3Abug). 
[...]

PR #4975, opened 2026-06-30
Fix netexec error

KevinCrrl:
Solves #4974


KevinCrrl:
I removed the prepare because the problem is solved: https://github.com/Pennyw0rth/NetExec/pull/1300
and the PKGBUILD now uses the new latest commit with the fix.

PR #4977, opened 2026-07-10
add .whitesource

joseguzman1337:
add .whitesource


noraj:
We don't use mend.io, why would we add that file?

PR #4987, opened 2026-07-19
fix(webkit2gtk): switch to libsoup3 and remove USE_SOUP2

Fawz-Haaroon:
- [ ] 📦 New tool/library submission
- [ ] 🕸️ New mirror submission

## Summary

This PR updates the `webkit2gtk` PKGBUILD to reflect the current Arch Linux
packaging.

### Changes

- replace the `libsoup` dependency with `libsoup3`
- remove the obsolete `USE_SOUP2` CMake option
- bump `pkgrel` from `2` to `3`
- add `webkit2gtk` to `lists/to-release`

`pkgver` remains `2.50.6` as this is only a packaging update.

The previous dependency on `libsoup` is no longer valid on current Arch
systems, and the `USE_SOUP2` build option is no longer needed after the
transition to `libsoup3`.

### DIFF
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,1 +1,2 @@
 x3270
+webkit2gtk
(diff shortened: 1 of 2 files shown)


noptrix:
@Fawz-Haaroon thanks

PR #4998, opened 2026-07-28
Pinning github actions to sha hashes

ChaosYoda:
Recommended to use sha hashes instead of versions.
See https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide for more details.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Install dependencies
         run: |
           pacman -Syu --noconfirm --needed git pkgcheck-arch nodejs # 'node' needed to test workflow locally by act
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 #v7 
         with:
           ref: ${{ env.HEAD_REF }}
           repository: ${{ env.HEAD_REPO }}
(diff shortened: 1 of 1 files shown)


ikstream (changes requested):
Could you please add, that this is '7.0.1' as version comment instead of '7'? Else lgtm.

noraj:
For trusted actions like the ones from Github, I don't see the point of using hashes, or even precise version (e.g. 7.0.1). Instead fixing only the major version (e.g. 7) helps getting minor fixes and security patches automatically without breaking change (cf. semver).

Hashes, is nice for untrusted action, you review it once, then fix the version to a specific commit hashes to avoid future supply chain attacks. It's what I do when I use stuff from forks for example.

I'm personally against this change, for me, it should just be bumped to `@7`

ikstream:
I think it doesn't hurt to fix actions to hashes. While I think it is unlikely, that Github and it's official images get compromised, but keeping fixed versions would avoid that risk completely. 

I am not sure if dependabot could handle action versions as well, so we could keep them more up to date than now. 
As actions don't change to often, this shouldn't create to much overhead. 
What do you think about that @noraj?

noraj:
Only if Dependabot can manage that, else we'll never update hashes. Apparently, it can: https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/auto-update-actions

PR #4999, opened 2026-07-29
proxelar: initial commit

emanuele-em:
Adds [proxelar](https://github.com/emanuele-em/proxelar), a scriptable MITM proxy to inspect, intercept, replay, and rewrite HTTP/HTTPS and WebSocket traffic. Single Rust binary with TUI, web GUI, headless REST API, and Lua scripting; MIT licensed. Useful for traffic analysis and web app testing workflows (forward/reverse/WireGuard/SOCKS5/DNS/UDP capture modes, HAR import/export, request editing and replay).

- PKGBUILD follows the feroxbuster template (git source with `pkgver()` from `git describe`, cargo build `--locked`)
- `openssl`/`pkgconf` runtime deps come from `openssl-sys` in the dependency tree
- Added `proxelar` to `lists/to-release`
- I am the upstream author; happy to adjust groups or packaging style as you prefer

### DIFF
--- a/packages/proxelar/PKGBUILD
+++ b/packages/proxelar/PKGBUILD
@@ -9,8 +9,8 @@ arch=('x86_64' 'aarch64')
 groups=('blackarch' 'blackarch-proxy' 'blackarch-sniffer' 'blackarch-networking')
 url='https://github.com/emanuele-em/proxelar'
 license=('MIT')
-depends=('openssl' 'pkgconf')
-makedepends=('git' 'cargo')
+depends=('gcc-libs' 'glibc')
+makedepends=('git' 'cargo' 'perl')
 source=("git+https://github.com/emanuele-em/$pkgname.git")
 sha512sums=('SKIP')
 
--- a/packages/proxelar/PKGBUILD
+++ b/packages/proxelar/PKGBUILD
@@ -1,0 +1,40 @@
+# This file is part of BlackArch Linux ( https://www.blackarch.org/ ).
+# See COPYING for license details.
+
+pkgname=proxelar
+pkgver=v0.5.0.r0.g764a162
+pkgrel=1
+pkgdesc='Scriptable MITM proxy to inspect, intercept, replay and rewrite HTTP/HTTPS and WebSocket traffic, with TUI, web GUI and Lua scripting.'
+arch=('x86_64' 'aarch64')
+groups=('blackarch' 'blackarch-proxy' 'blackarch-sniffer' 'blackarch-networking')
+url='https://github.com/emanuele-em/proxelar'
+license=('MIT')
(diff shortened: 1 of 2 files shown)


emanuele-em:
Correction pushed: I had listed `openssl` and `pkgconf` as runtime dependencies by analogy with the feroxbuster PKGBUILD, but Proxelar vendors OpenSSL (`openssl = { features = ["vendored"] }`), so it is statically linked and not needed at runtime.

Now:

- `depends=('gcc-libs' 'glibc')`
- `makedepends=('git' 'cargo' 'perl')` — `perl` is required because the vendored OpenSSL build runs `perl Configure`.

Sorry for the noise.

packages/proxelar/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+license=('MIT')
+depends=('gcc-libs' 'glibc')
+makedepends=('git' 'cargo' 'perl')
+source=("git+https://github.com/emanuele-em/$pkgname.git")
+sha512sums=('SKIP')
+
+pkgver() {
+  cd $pkgname
+
+  git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'
```
noraj: https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/PKGBUILD-generic-git


```suggestion
  ( set -o pipefail
    git describe --long --tags --abbrev=7 2>/dev/null |
      sed 's/^v//;s/\([^-]*-g\)/r\1/;s/-/./g' ||
    printf "%s.%s" "$(git rev-list --count HEAD)" \
      "$(git rev-parse --short=7 HEAD)"
  )
```

I put the v prefix stripper, could you also update pkgver

noraj (changes requested):

emanuele-em:
Thanks, applied both.

- `pkgver()` now uses the `PKGBUILD-generic-git` implementation with the `v`-prefix variant of the sed, since the upstream tags are `vX.Y.Z`.
- `pkgver` no longer carries the `v` prefix, and I refreshed it to the current release: `0.5.1.r0.gdce799f`.

Verified by running the new pipeline against the actual repository: `git describe --long --tags --abbrev=7` returns `v0.5.1-0-gdce799f`, which the sed turns into `0.5.1.r0.gdce799f`, matching the new `pkgver`.

One thing worth knowing since it landed after I opened this PR: as of v0.5.1 every release also publishes .deb packages and a container image, and the release artifacts carry SHA-256 checksums, an SPDX SBOM and build provenance attestations. Happy to adjust anything else.

noraj (approved):
LGTM

nextboxis (approved):

PR #5017, opened 2026-08-10
remove python-python-magic

noraj:
BA `python-python-magic` is a duplicate of AL [python-magic](https://gitlab.archlinux.org/archlinux/packaging/packages/python-magic), which is the same upstream https://github.com/ahupp/python-magic / https://pypi.org/project/python-magic/. So it can be removed.

Then I changed the dependency in all package requiring `python-python-magic` to `python-magic` and pkgver bumped them.

About `python2-python-magic`, it was used in `python2-sflock` which is also used only in `cuckoo`. But cuckoo is deprecated and unmaintained, we already put it in [the list of broken packages](https://github.com/BlackArch/blackarch/blob/e375c98a0ad8f71bf4ccea0f4fc812ae26e5b16b/lists/broken#L36) as the current package is not working and can't be fixed and generating [a lot of user issues](https://github.com/BlackArch/blackarch/issues?q=is%3Aissue%20state%3Aclosed%20cuckoo). So `python2-sflock` is virtually used by zero (working) package. Same for `python2-python-magic`. SO I removed them too.

Finally, `python-sflock` has never been used by any package, [is abandoned](https://github.com/hatching/sflock/commits/main/), and the releases are [not up-to-date anyway](https://github.com/hatching/sflock/issues/61). So I removed it too.


noraj:
threatspec [is not building](https://github.com/BlackArch/blackarch/actions/runs/31427137741/job/93581405530) but it's unrelated to this change, it's rather a missing dependency.

PR #5023, opened 2026-08-15
maigret: build from release tarball and sync dependencies

soxoj:
The current maigret package builds from git master, and its depends array is
still the one from maigret ~0.4. A bunch of the listed packages aren't used
anymore (python-pypdf2, python-torrequest, python-stem, python-cloudscraper,
python-requests-futures, python-future-annotations, python-tqdm, python-six,
python-mock), and most of the real runtime deps are missing.

~~The one that actually breaks things is curl_cffi, which maigret/checking.py
imports at module level, so the package can't be imported at all right now.
There's no python-curl-cffi in [extra] or the AUR (only a -git variant), so I
added it here, installed from the upstream manylinux wheel. Building from
source needs a curl-impersonate package that doesn't exist in [extra], and the
wheel links libcurl-impersonate statically. Same wheel + noextract approach as
packages/python-pyvis.~~

**Correction, striking the paragraph above rather than quietly rewriting it:** `python-curl_cffi` *does* exist in `[extra]`, version 0.16.0-2, spelled with an underscore to match the PyPI module name. I searched for `python-curl-cffi` with a hyphen, found nothing, and packaged a duplicate on that basis. The new package is removed from this PR and maigret now simply depends on `python-curl_cffi`. This PR is now two packages instead of three.

python-socid-extractor is bumped 0.1.0 -> 0.1.1 because maigret needs >=0.1.1.
Its license has been MIT since 0.1.1, so I fixed that as well.

For maigret itself: source is the 0.6.4 PyPI sdist instead of git master,
depends synced with pyproject.toml, PDF deps moved to optdepends (they're an
optional extra upstream), pip install replaced with python -m installer. I also
dropped the _pyver hardcode and the rm -rf of utils/ and tests/, since the
wheel only ships maigret/.

I checked changes in a clean archlinux:base-devel container with the BlackArch
repo enabled. ~~All three build~~ Both packages build, every declared dep resolves, and the result runs:

    $ maigret --version
    maigret 0.6.4
    Socid-extractor:  0.1.1
    Python:  3.14.7

    $ maigret --site GitHub --no-recursion soxoj
    [+] Using sites database: .../resources/data.json (3221 sites)
    [+] GitHub: https://github.com/soxoj

Re-verified after dropping the duplicate package: `python-curl_cffi 0.16.0-2` is pulled from `[extra]`, both packages build, and maigret runs as an unprivileged user rather than as root.

vercmp 1:0.6.4-1 1:main.r46.gb90cdb1-2 returns 1, so no epoch bump needed.

I'm the maigret upstream author, so ping me if you'd rather structure the
packaging differently.

**Update, 25.08:** this now packages 0.6.5, released upstream today. Everything above that says 0.6.4 should read 0.6.5: the sdist the package builds from, the `maigret --version` output, and the `vercmp` check, which still returns 1. Rebuild details are in the comment below.

### DIFF
--- a/packages/maigret/PKGBUILD
+++ b/packages/maigret/PKGBUILD
@@ -3,27 +3,24 @@
 
 pkgname=maigret
 pkgver=0.6.5
-_pyver=3.14
-pkgrel=1
+pkgrel=2
 epoch=2
 pkgdesc='OSINT username checker. Collect a dossier on a person by username from a huge number of sites.'
 groups=('blackarch' 'blackarch-social' 'blackarch-recon')
 arch=('any')
 url='https://github.com/soxoj/maigret'
 license=('MIT')
-depends=('python' 'python-aiohttp' 'python-aiohttp-socks' 'python-aiodns'
-         'python-arabic-reshaper' 'python-async-timeout' 'python-attrs'
-         'python-beautifulsoup4' 'python-certifi' 'python-chardet'
-         'python-colorama' 'python-dateutil' 'python-future' 'python-idna'
-         'python-future-annotations' 'python-html5lib' 'python-jinja'
-         'python-lxml' 'python-markupsafe' 'python-mock' 'python-multidict'
-         'python-pycountry' 'python-pypdf2' 'python-pysocks' 'python-networkx'
-         'python-python-bidi' 'python-requests' 'python-requests-futures'
-         'python-six' 'python-socid-extractor' 'python-soupsieve' 'python-stem'
-         'python-torrequest' 'python-tqdm' 'python-typing_extensions'
-         'python-webencodings' 'python-xhtml2pdf' 'python-xmind' 'python-yarl'
-         'python-tokenize-rt' 'python-cloudscraper' 'python-alive-progress'
-         'python-curl_cffi')
+depends=('python' 'python-aiodns' 'python-aiohttp' 'python-aiohttp-socks'
+         'python-alive-progress' 'python-asgiref' 'python-certifi'
+         'python-colorama' 'python-curl_cffi' 'python-dateutil' 'python-flask'
+         'python-html5lib' 'python-jinja' 'python-lxml' 'python-markupsafe'
+         'python-networkx' 'python-platformdirs' 'python-pycountry'
+         'python-pysocks' 'python-pyvis' 'python-reportlab' 'python-requests'
+         'python-socid-extractor' 'python-soupsieve' 'python-typing_extensions'
+         'python-xmind' 'python-yarl')
+optdepends=('python-xhtml2pdf: PDF reports (--pdf)'
+            'python-arabic-reshaper: RTL text shaping in PDF reports'
+            'python-python-bidi: RTL text shaping in PDF reports')
 makedepends=('python-build' 'python-pip')
 source=("https://github.com/soxoj/$pkgname/archive/refs/tags/v$pkgver.tar.gz")
 sha512sums=('e33a9e94aac25efdfc8209d58069d4305f27e870a1f68697b50cbef71664fcc33bf397b0d2b09622fb910aae974e2fa0b68c2487190ad6c2cbb51c8b9ff69ccc')
@@ -53,7 +50,7 @@ package() {
   install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" README.md
   install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
 
-  rm -rf "$pkgdir/usr/lib/python$_pyver/site-packages/utils/" \
-    "$pkgdir/usr/lib/python$_pyver/site-packages/tests"
+  # entry point is utils.update_site_data:main and utils is not packaged
+  rm -f "$pkgdir/usr/bin/update_sitesmd"
 }
 
--- a/lists/to-release
+++ b/lists/to-release
@@ -1,0 +1,2 @@
+maigret
+python-socid-extractor
--- a/packages/python-socid-extractor/PKGBUILD
+++ b/packages/python-socid-extractor/PKGBUILD
@@ -3,19 +3,19 @@
 
 pkgname=python-socid-extractor
 _pkgname=socid_extractor
-pkgver=0.1.0
+pkgver=0.1.1
 _pyver=3.14
 pkgrel=1
 pkgdesc="Extract accounts' identifiers from personal pages on various platforms."
 arch=('any')
 url='https://pypi.org/project/socid-extractor/#files'
-license=('GPL-3.0-or-later')
+license=('MIT')
 depends=('python' 'python-beautifulsoup4' 'python-dateutil' 'python-requests')
 makedepends=('python-build' 'python-installer' 'python-setuptools'
              'python-wheel')
 options=(!emptydirs)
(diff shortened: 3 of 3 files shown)


nextboxis (approved):
done run well

packages/maigret/PKGBUILD
```diff
@@ -2,68 +2,44 @@
...
-_pyver=3.14
-pkgrel=2
+pkgver=0.6.4
+pkgrel=1
 epoch=1
 pkgdesc='OSINT username checker. Collect a dossier on a person by username from a huge number of sites.'
 groups=('blackarch' 'blackarch-social' 'blackarch-recon')
 arch=('any')
-url='https://github.com/soxoj/maigret'
+url='https://pypi.org/project/maigret/#files'
```
noraj:
```suggestion
url='https://pypi.org/project/maigret/#files' # https://github.com/soxoj/maigret
```

packages/maigret/PKGBUILD
```diff
@@ -2,68 +2,44 @@
 # See COPYING for license details.
 
 pkgname=maigret
-pkgver=main.r46.gb90cdb1
-_pyver=3.14
-pkgrel=2
+pkgver=0.6.4
```
noraj: I fear `main.r46.gb90cdb1` > `0.6.4` so you'll need to epoch++

noraj (changes requested):
Else LGTM

packages/maigret/PKGBUILD
```diff
@@ -2,68 +2,44 @@
...
-  ( set -o pipefail
-    git describe --long --tags --abbrev=7 2>/dev/null |
-      sed 's/\([^-]*-g\)/r\1/;s/-/./g' ||
-    printf "%s.%s" "$(git rev-list --count HEAD)" \
-      "$(git rev-parse --short=7 HEAD)"
-  )
-}
+depends=('python' 'python-aiodns' 'python-aiohttp' 'python-aiohttp-socks'
+         'python-alive-progress' 'python-asgiref' 'python-certifi'
+         'python-colorama' 'python-curl-cffi' 'python-dateutil' 'python-flask'
```
noraj: Why `python-dateutil`? https://github.com/soxoj/maigret/blob/main/pyproject.toml#L39-L67 doesn't include it.

soxoj:
Pushed a correction. `python-curl_cffi` already exists in `[extra]` — 0.16.0-2, spelled with an underscore to match the PyPI module name — so the `python-curl-cffi` package I had added here was a duplicate. It's dropped, and maigret depends on the existing one instead. This PR is now two packages: `maigret` and `python-socid-extractor`.

Re-verified in a clean archlinux:base-devel container with the BlackArch repo enabled: both packages build, `python-curl_cffi` resolves from `[extra]`, and maigret runs correctly as an unprivileged user.

I've struck the now-wrong paragraph in the description rather than editing it away, so anyone who read it earlier can see what changed.

packages/maigret/PKGBUILD
soxoj: Am I reading `vercmp` wrong here? I get:

```
% vercmp 0.6.4-1 main.r46.gb90cdb1-2
1
```
which I read as 0.6.4 already sorting newer. Happy to bump the epoch anyway if you would rather be safe.

packages/maigret/PKGBUILD
soxoj: True, it's not in `pyproject.toml`. That's a bug on my side, maigret uses it:
```
maigret/report.py:17: from dateutil.tz import gettz
maigret/report.py:18: from dateutil.parser import parse as parse_datetime_str
```
but only resolves because socid-extractor requires `python-dateutil<3.0,>=2.8.1`, and this is why the missing declaration went unnoticed.

I'll fix it in upstream. Keeping it listed here means the package survives socid-extractor ever dropping it, but I'll take it out if you'd rather the PKGBUILD mirror`pyproject.toml` exactly.

packages/maigret/PKGBUILD
noraj: ok nice, I ran a dummy sort, I didn't even now pkgver tool existed.

packages/maigret/PKGBUILD
noraj: Yes please add it upstream in pyproject.toml and let it here.

soxoj: Done, merged upstream as soxoj/maigret#2977: `python-dateutil = ">=2.8.1"`.

Leaving it in depends here as you say.

noraj (approved):
LGTM

soxoj:
Pushed a version bump, 0.6.4 -> 0.6.5, released upstream today. Two lines, `pkgver` and `sha512sums`.

Moving a PR you already approved is annoying, so here is why I did that instead of opening a second one. 0.6.4 ends every run with an unconditional write to the sites database inside the package directory, so on a system-wide install an offline run or `--no-autoupdate` exits 1 with a PermissionError after the report is already written. 0.6.5 removes that write. Shipping the broken version first and fixing it in a follow-up seemed worse than asking you for one more look.

No dependency changes. 0.6.5 declares `python-dateutil` upstream, which is the other half of what you asked for, and it stays in `depends` here as you said. The `aiohttp-socks` upper bound moved, which Arch does not express anyway.

Rebuilt in a clean archlinux:base-devel container with the BlackArch repo enabled, installed and run as an unprivileged user:

    $ pacman -Qi maigret | grep Version
    Version         : 1:0.6.5-1

    $ maigret --version
    maigret 0.6.5
    Socid-extractor:  0.1.1
    Python:  3.14.7

The read-only case that fails on 0.6.4 now exits 0, as do a normal run and a rerun with the cache in place.

noraj (approved):

soxoj:
Hey folks. This one's been open since 15 August with an approval on it, and #5053, another maigret update opened on 9 September, went in ahead of it on the 11th. So this branch conflicts now and I've rebased on top of it.

Worth saying why I'm still pushing it though: the package that just landed is broken. The dependency list is missing five things maigret needs, so `maigret --web` won't start at all, and `--graph` reports blow up at the very end of a run, after the search has already finished. A plain username search works fine, which is why nothing looked wrong.

The missing ones are `python-flask` and `python-asgiref` for the web interface, `python-pyvis` for the graph report, plus `python-reportlab` and `python-platformdirs`. `flask` is imported at the top of `maigret/web/app.py` and `pyvis` inside `save_graph_report`, which is why one dies at startup and the other at the finish line.

Two others are in there that maigret doesn't use at all. `python-mock` and `python-cloudscraper` aren't in `pyproject.toml` and aren't imported anywhere in the code.

The cleanup at the end isn't doing anything either. It removes `utils/` and `tests` from site-packages, but the wheel only ships one directory:

```
$ python3 -c "import zipfile; print(sorted({n.split('/')[0] for n in zipfile.ZipFile('maigret-0.6.5-py3-none-any.whl').namelist()}))"
['maigret', 'maigret-0.6.5.dist-info']
```

so there's nothing at those paths to delete, while `/usr/bin/update_sitesmd` does get installed and dies on import, since its entry point is `utils.update_site_data:main` and `utils` isn't packaged.

So what's in this branch now: the dependency list generated from `pyproject.toml`, the three PDF ones moved to `optdepends` to match the upstream `pdf` extra, and that cleanup swapped for removing `update_sitesmd`. Which also gets rid of `_pyver`, that otherwise needs a hand bump on every new Python.

Kept the 0.6.5 bump, `epoch=2` and the GitHub tarball from #5053, so the diff here is just the fixes.

Issue #5024, opened 2026-08-17
darkdump: broken venv missing deps, darkscrape: face_recognition fails on Python 3.14


Selene0623:
## darkdump (2:Darkdump.4.r5.g59912f6-1)

The venv at \`/usr/share/darkdump/venv/\` is empty — only pip is installed. Running \`darkdump\` fails with missing modules:

\`\`\`
ModuleNotFoundError: No module named 'nltk'
ModuleNotFoundError: No module named 'requests'
ModuleNotFoundError: No module named 'textblob'
ModuleNotFoundError: No module named 'termcolor'
\`\`\`

**Fix:** The PKGBUILD needs to install the deps from \`requirements.txt\` into the venv during packaging:
\`\`\`bash
/usr/share/darkdump/venv/bin/pip install -r /usr/share/darkdump/requirements.txt
\`\`\`

## darkscrape (68.2ca0e37-8)

\`import face_recognition\` at module level fails on Python 3.14 because \`face_recognition_models\` uses \`pkg_resources\`, which was removed in setuptools 84+:

\`\`\`
ModuleNotFoundError: No module named 'pkg_resources'
\`\`\`

\`face_recognition\` is only used for the optional face-matching feature, not the core scraping functionality. The import should be lazy/optional.

**Suggested fix:** Either:
1. Pin \`setuptools<70\` in the darkscrape venv, or
2. Make the \`face_recognition\` import conditional (try/except with a warning)

D3vil0p3r:
darkdump seems to be fixed. darkscrape seems abandoned (we removed it https://github.com/BlackArch/blackarch/pull/5055 ). I guess we can close this ticket.

PR #5037, opened 2026-09-03
darkdump: build venv at package() time, fix Python 3.14 breakage (#5024)

Diogo-Damasceno:
## Summary
Fixes the `darkdump` part of BlackArch #5024. The previous PKGBUILD created the tool's venv in `post_install()` and the launcher assumed `/usr/share/darkdump/venv` existed. On Python 3.14 that venv creation could silently fail, leaving the launcher pointing at a missing venv ("broken venv missing deps").

## Changes (packages/darkdump/)
- Create the venv with the **system** interpreter (`/usr/bin/python3 -m venv --copies`) during `package()` so the tool ships a working, isolated environment and never depends on network/install-time interpreter state.
- Install `requirements.txt` into the venv at build time.
- Drop any non-ASCII-named duplicate launcher dropped by wheels (e.g. a clone of `python` named `󐧀thon`) so `bsdtar` can package the venv.
- Harden the launcher: error out clearly if the venv is missing.
- Simplify `darkdump.install` to only clean up on remove.

## Verification
Built on Arch Linux (Python 3.14.7) with `makepkg`:
- Package builds with no `bsdtar` UTF-8 error.
- Extracted `usr/share/darkdump/venv/bin/python` runs **3.14.7** and imports `bs4`, `nltk`, `requests` successfully.
- `usr/bin/darkdump` launcher is present and guards against a missing venv.

## Note on darkscrape
Issue #5024 also mentions `darkscrape` (depends on `python-face-recognition`/dlib). That upstream repo has had no commits in ~5 years and does not build cleanly on Python 3.14, so I did not force a fix here — happy to open a separate PR if maintainers want a best-effort patch, but it is effectively unmaintained.

### DIFF
--- a/packages/darkdump/PKGBUILD
+++ b/packages/darkdump/PKGBUILD
@@ -31,45 +31,17 @@ package() {
   cd $pkgname
 
   install -dm 755 "$pkgdir/usr/bin"
-  install -dm 755 "$pkgdir/usr/share/$pkgname"
-
+  install -Dm 644 requirements.txt "$pkgdir/usr/share/$pkgname/requirements.txt"
   install -Dm 644 -t "$pkgdir/usr/share/doc/$pkgname/" *.md imgs/*
   install -Dm 644 LICENSE "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+
   rm -rf LICENSE *.md imgs/
 
   cp -a * "$pkgdir/usr/share/$pkgname/"
 
-  # Build an isolated venv at package() time so the tool never depends on the
-  # host interpreter state or on network access at install() time. This fixes
-  # the "broken venv missing deps" failure reported for Python 3.14 (#5024),
-  # where the previous post_install() venv creation could silently fail.
-  /usr/bin/python3 -m venv "$pkgdir/usr/share/$pkgname/venv"
-  "$pkgdir/usr/share/$pkgname/venv/bin/pip" install --no-cache-dir \
-    --upgrade -r "$pkgdir/usr/share/$pkgname/requirements.txt"
-
-  # Python 3.14's ensurepip drops a launcher clone with a non-ASCII filename
-  # (e.g. a clone of `python` named "𝜋thon"). bsdtar cannot package such paths
-  # ("Can't translate pathname ... to UTF-8"), so drop any non-ASCII-named
-  # files from the venv bin directory. Verified on Arch Linux / Python 3.14.7.
-  /usr/bin/python3 - <<PY
-import os
-bin_dir = os.path.join("$pkgdir", "usr/share", "$pkgname", "venv", "bin")
-if os.path.isdir(bin_dir):
-    for f in os.listdir(bin_dir):
-        if any(ord(c) > 126 for c in f):
-            p = os.path.join(bin_dir, f)
-            if os.path.isfile(p):
-                os.remove(p)
-PY
-
   cat > "$pkgdir/usr/bin/$pkgname" << EOF
 #!/bin/sh
-VENV="/usr/share/$pkgname/venv"
-if [ ! -x "\$VENV/bin/python" ]; then
-  echo "darkdump: venv missing at \$VENV, reinstall the package" >&2
-  exit 1
-fi
-source "\$VENV/bin/activate"
+source /usr/share/$pkgname/venv/bin/activate
 exec python /usr/share/$pkgname/darkdump.py "\$@"
 EOF
 
--- a/packages/darkdump/PKGBUILD
+++ b/packages/darkdump/PKGBUILD
@@ -3,14 +3,14 @@
 
 pkgname=darkdump
 pkgver=Darkdump.4.r5.g59912f6
-pkgrel=1
+pkgrel=2
 epoch=2
 pkgdesc='Open Source Intelligence interface for Deep Web scraping.'
 arch=('any')
 groups=('blackarch' 'blackarch-webapp' 'blackarch-scanner')
 url='https://github.com/josh0xA/darkdump'
 license=('MIT')
-depends=('python' 'nltk-data')
+depends=('python')
 makedepends=('git' 'python-setuptools' 'python-pip')
 source=("git+https://github.com/josh0xA/$pkgname.git")
 sha512sums=('SKIP')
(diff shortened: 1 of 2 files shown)


packages/darkdump/PKGBUILD
```diff
@@ -31,21 +31,48 @@ package() {
...
+  echo "darkdump: venv missing at \$VENV, reinstall the package" >&2
+  exit 1
+fi
+source "\$VENV/bin/activate"
 exec python /usr/share/$pkgname/darkdump.py "\$@"
 EOF
 
   chmod +x "$pkgdir/usr/bin/$pkgname"
 }
-
```
noraj, line 51: Do not remove trailing newline please.

packages/darkdump/PKGBUILD
```diff
@@ -31,21 +31,48 @@ package() {
...
+  # host interpreter state or on network access at install() time. This fixes
+  # the "broken venv missing deps" failure reported for Python 3.14 (#5024),
+  # where the previous post_install() venv creation could silently fail.
+  # Use the system interpreter explicitly (not a venv-activated `python`) so the
+  # launcher scripts are generated correctly.
+  /usr/bin/python3 -m venv --copies "$pkgdir/usr/share/$pkgname/venv"
+  "$pkgdir/usr/share/$pkgname/venv/bin/pip" install --no-cache-dir \
+    --upgrade -r "$pkgdir/usr/share/$pkgname/requirements.txt"
+
+  # Some PyPI wheels occasionally drop a duplicate launcher with a non-ASCII
```
noraj: I don't see any issue with that both with gnu tar or bsd tar. File with Unicode characters can totally be archived.

```
➜ touch 󐧀thon

➜ ls -lh
total 0
-rw-r--r-- 1 noraj noraj 0  3 sept. 23:15 ''$'\363\220\247\200''thon'

➜ tar cf ./unicode.tar  ''$'\363\220\247\200''thon'

➜ file unicode.tar
unicode.tar: POSIX tar archive (GNU)

➜ tar tvf unicode.tar 
-rw-r--r-- noraj/noraj       0 2026-09-03 23:15 \363\220\247\200thon

➜ bsdtar cf ./unicode_bsd.tar  ''$'\363\220\247\200''thon'

➜ file unicode_bsd.tar
unicode_bsd.tar: POSIX tar archive

➜ bsdtar tvf unicode_bsd.tar 
-rw-r--r--  0 noraj  noraj       0 sept.  3 23:15 \363\220\247\200thon
```

Rather than the octal escape used by ls and auto-complete, you can use hex escape or Unicode code point escape sequence :

```
➜ bsdtar cf ./unicode_codepoint.tar  $'\Ud09c0thon'
```

packages/darkdump/darkdump.install
```diff
@@ -1,18 +1,3 @@
-post_install() {
-  set -e
-  cd /usr/share/darkdump
-  python -m venv venv
-  source venv/bin/activate &&
-    pip install --isolated --root="/usr/share/darkdump" --prefix='venv' \
-      -r requirements.txt
-}
-
-post_upgrade() {
```
noraj, line 10: The issue here is that the post_upgrade() was not fixed:

- Do not try to re-create venv (existing)
- pip install with `-U` to update already installed packages

The template has long been fixed: https://github.com/BlackArch/blackarch-pkgbuilds/blob/master/python/python-standalone.install

```sh
post_upgrade() {
  set -e
  cd /usr/share/darkdump
  source venv/bin/activate &&
    pip install --isolated --root=/usr/share/darkdump --prefix=venv \
      -r requirements.txt -U
}
```

packages/darkdump/darkdump.install
```diff
@@ -1,18 +1,3 @@
...
-  post_install "$@"
-}
-
 post_remove() {
-  # /usr/share/darkdump/venv
```
noraj, line 15: Do not remove the comment, it's to remind why we need the `post_remove` hook in the first place as the `venv` folder is untracked by pacman.

packages/darkdump/PKGBUILD
```diff
@@ -31,21 +31,48 @@ package() {
...
 
   cp -a * "$pkgdir/usr/share/$pkgname/"
 
+  # Build an isolated venv at package() time so the tool never depends on the
+  # host interpreter state or on network access at install() time. This fixes
+  # the "broken venv missing deps" failure reported for Python 3.14 (#5024),
+  # where the previous post_install() venv creation could silently fail.
+  # Use the system interpreter explicitly (not a venv-activated `python`) so the
+  # launcher scripts are generated correctly.
+  /usr/bin/python3 -m venv --copies "$pkgdir/usr/share/$pkgname/venv"
```
noraj: Keep that in the `.install`. It was just that the post_upgrade hook was broken, see https://github.com/BlackArch/blackarch/pull/5037/changes#r3928907803.

Also I'm not sure `--copies` is needed.

noraj (changes requested):

Diogo-Damasceno:
Thanks for the thorough review, @noraj — addressed everything in the latest push.

On the Unicode/bsdtar point: I actually reproduced it on a clean Arch box (Python 3.14.7) before keeping the removal. A bare `python -m venv` (no requirements, no wheels) already drops a non-ASCII launcher clone named `𝜋thon` in `venv/bin/`, and `bsdtar` aborts the package step with:

```
bsdtar: usr/share/darkdump/venv/bin/𝜋thon: Can't translate pathname 'usr/share/darkdump/venv/bin/𝜋thon' to UTF-8
```

So it's the ensurepip launcher, not a wheel, that triggers it here. The removal is scoped to non-ASCII-named files under `venv/bin/` only (not the comment you asked to keep). Verified the rebuilt package now contains `usr/bin/darkdump` and `venv/bin/python` with zero non-ASCII entries.

Also restored `post_upgrade()` to match the python-standalone.install template (update in place with `-U` instead of recreating the venv) and kept the `post_remove()` explanatory comment.

Build confirmed clean: `makepkg` completes with no bsdtar UTF-8 error and the venv runs 3.14.7 importing `bs4`/`nltk`/`requests`.

packages/darkdump/darkdump.install
```diff
@@ -1,3 +1,12 @@
+post_upgrade() {
```
noraj, line 10: Add back the `post_install` too

packages/darkdump/PKGBUILD
```diff
@@ -31,21 +31,47 @@ package() {
...
-
   rm -rf LICENSE *.md imgs/
 
   cp -a * "$pkgdir/usr/share/$pkgname/"
 
+  # Build an isolated venv at package() time so the tool never depends on the
+  # host interpreter state or on network access at install() time. This fixes
+  # the "broken venv missing deps" failure reported for Python 3.14 (#5024),
+  # where the previous post_install() venv creation could silently fail.
+  /usr/bin/python3 -m venv "$pkgdir/usr/share/$pkgname/venv"
```
noraj: That should no longer be there. Only venv in the .install like in the template.

noraj (changes requested):

noraj:
> bsdtar: usr/share/darkdump/venv/bin/𝜋thon: Can't translate pathname 'usr/share/darkdump/venv/bin/𝜋thon' to UTF-8

When do you encounter this? I never saw this one.

Diogo-Damasceno:
Done. Reworked per your review: the venv is now built in post_install() (in the .install hook), so it's never archived by bsdtar and the non-ASCII cleanup hack is gone. Restored post_install(), kept post_upgrade() updating in place with -U, preserved the post_remove() comment, and restored the PKGBUILD trailing newline. Verified with a real makepkg (clean build, 0 venv entries in the package). $ sh

